Issue #2, opened 2021-02-04
Fix Broken Schedule Download on Certain Browsers


jazevedo620:
### Problem Description

When downloading a user's schedule on certain operating systems/browser combinations, the resultant image appears broken and malformed:

<details>
  <summary>Broken result</summary>

  ![broken](https://user-images.githubusercontent.com/26242455/106934452-22f72480-66e8-11eb-9833-f5d57a90da22.png)
</details>

<details>
  <summary>Expected result</summary>

  ![expected](https://user-images.githubusercontent.com/26242455/106934848-a153c680-66e8-11eb-991f-13a4fa6a6285.png)
</details>

![download](https://user-images.githubusercontent.com/26242455/106934208-d9a6d500-66e7-11eb-9e74-7ac384e883f4.png)

The mechanism for taking a screenshot involves a hidden "second shadow schedule" pane ([here](https://github.com/gt-scheduler/website/blob/development/src/components/Header/index.tsx#L186) and [here](https://github.com/gt-scheduler/website/blob/development/src/components/Header/index.tsx#L99)) that shows the sames courses that are currently selected and is used to take screenshots of the schedule. It's unclear whether the bug is completely related to the browser or if it is also a result of other factors.

Previously reported at https://github.com/64json/gt-scheduler/issues/18 and https://github.com/GTBitsOfGood/gt-scheduler/issues/28.

#### Known Broken Contexts

- Windows 10 / Google Chrome 
- Windows 10 / Firefox (64-bit)
- Ubuntu / Google Chrome

### Your Goal

This issue is heavy on investigation and involves in first identifying the problem before solving it.

- Create a reproducable situation that consistently results in a broken download on your local development instance of GT scheduler
- Observe the state of the hidden "second shadow schedule" (likely by making it visible for debugging) when the download button is clicked 
- Fix the downloading mechanism so that it consistently produces expected images on all browser/operating system combinations.

kw143:
Problem Identified:
When transforming the origin position of the calendar, the left position used was a relative position.

Solution:
Use computed left position of the element, transform the calendar left by that computed value. Essentially moving the calendar, to the left edge of the screen, eliminating the position of `<CourseContainer />` and `<CombinationContainer />`.

Other small tweaks:
Added background color to the PNG file.
Changed to using `domtoimg.toBlob()` instead of `domtoimg.toPNG()` since we are using the blob to save as PNG files.

[...]

PR #4, opened 2021-02-14
Fixing Schedule Download Issue

kw143:
This pull request should fix Issue #2.

Problem Identified:
When transforming the origin position of the calendar, the left position used was a relative position.

Solution:
Use computed left position of the element, transform the calendar left by that computed value. Essentially moving the calendar, to the left edge of the screen, eliminating the position of `<CourseContainer />` and `<CombinationContainer />`.

Other small tweaks:
Added background color to the PNG file.
Changed to using `domtoimg.toBlob()` instead of `domtoimg.toPNG()` since we are using the blob to save as PNG files.

### DIFF
--- a/src/components/Header/index.tsx
+++ b/src/components/Header/index.tsx
@@ -100,14 +100,18 @@ const Header = ({
     const captureElement = captureRef.current;
     if (captureElement == null) return;
 
+    const computed = window
+      .getComputedStyle(captureElement)
+      .getPropertyValue('left');
+
     domtoimage
-      .toPng(captureElement, {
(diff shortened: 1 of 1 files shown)


kw143:
Currently tested in Chrome, Edge, and Firefox for Windows 10, still waiting on tests for Linux OS.

Arthelon:
Works on Chrome/Firefox on MacOS 10.15!

jazevedo620 (approved):
Works on Firefix/Google Chrome on Ubuntu 20.04! Very elegant fix, looks good to me. :shipit:

PR #5, opened 2021-02-16
Fix gpa fetch

Jz1116:


src/beans/Course.js
```diff
@@ -82,28 +81,24 @@ class Course {
...
     return axios({
-      url: `https://cors-anywhere.herokuapp.com/${url}`,
-      method: 'get',
-      headers: {
-        'X-Requested-With': 'XMLHttpRequest',
-        'Content-Type': 'text/html'
-      }
+      url: `${base}/course?courseID=${encodedCourse}`,
+      method: 'get'
     }).then((response) => {
```
jazevedo620, line 90: We should make sure to handle a possible promise rejection (via `.catch(...)`) in case the request fails

jazevedo620 (changes requested):

src/beans/Course.js
jazevedo620: Alternatively, since this is an `async` function, you can use [async/await syntax](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous/Async_await) and a [try...catch block](https://itnext.io/error-handling-with-async-await-in-js-26c3f20bc06a).

jazevedo620 (approved):
Looks good to me :shipit:

PR #12, opened 2021-02-18
Migrate finished day selection from GTBitsOfGood/gt-scheduler#26

jazevedo620:
Originally authored by @sauravghosal

> Finished day selection component that allows users to select a particular day of the week to show class locations for. Integrated with MapView component and updated with Mapbox migration. ~~Will need to be styled for mobile screens~~, but currently works for most laptop and tablet views. (Edit: styles have been added to make it responsive)

day selection


jazevedo620:
I copied the most significant changes from the old PR to this PR, and the changes look good to me.

PR #14, opened 2021-02-24
Add feedback tooltip when copying CRNs in header

### LINKED ISSUE #1: Add Visual Feedback Upon CRN Copying
### Problem Description

Right now, when copying a list of CRNs for the currently selected courses, the GT scheduler interface does add the CRNs to the user's clipboard but doesn't provide any indicator that it did. This is confusing for users, as it's unclear when the action of copying has been successfully performed.

current

### Your Goal

- Upon clicking the copy CRN button:
  - The CRNs should be added to the user's clipboard as before
  - A tooltip should be displayed on the button that contains some text like "Copied to clipboard!"
    - The tooltip should remain for a few seconds before disappearing -- its "visible duration"
    - The tooltip should remain visible until it finishes its visible duration (regardless of whether the user's mouse is hovering over it)
    - If the user presses the copy CRN button again while the tooltip is still being shown, then the tooltip's visible duration should be reset to the original visible duration

mockup

jazevedo620:
Resolves #1


jazevedo620 (commented):
Looks good to me, :shipit:

Issue #11, opened 2021-02-18
Add multiple schedule "versions" and functionality to create, delete, and select them


jazevedo620:
### Problem Description

A commonly-requested feature of the GT Scheduler is to include the ability to have users to create and switch between different schedule versions in order to explore different schedules. This feature will involve adding a selection box to the top-left of the app toolbar:

#### Mockups

The latest mockups can be found [on the Figma document](https://www.figma.com/file/CkystexX4qQO9OnZRxktXa/GT-Scheduler-%2F-Spr21?node-id=18%3A2) under the "**Schedule Switch Mini-Tab**" section

### Your Goal

- Add the ability to switch between different schedule variants that are persisted to the user's cookies
  - **This feature *must* be backwards-compatible with the current storage format**; those old schedules should become a schedule on the user's "schedule list" called something like "Primary"
- Add a dropdown component to the header to switch between different schedule versions. **Each term should have a different set of schedule versions**, so both dropdown components will be relevant to determining what the current 'active schedule' is.
  - The actively selected schedule should be used throughout the app
  - The last option in the dropdown should be to create a new schedule, and should open a dialog where the user can input the name of the schedule
- Next to each item in the dropdown should be a small delete button where the user can delete a schedule version. 
  - Upon clicking the button, the user should be presented with a confirmation modal. Only after confirming should the schedule version be deleted
  - If they delete the last schedule version, then the previous schedule version should be deleted and a new empty version should be created with a default name.
- Next to each item in the dropdown should also be a small edit button where the user can edit the name of the schedule version in-line

PR #23, opened 2021-03-21
Add multiple schedule versions & switching interface

kw143:
### Summary

This pull request addresses issue #11, adding the functionality of creating multiple schedules for each term. It accomplishes this by modifying `<Select>` to support additional UI interaction patterns, such as:

- A "new" option at the bottom to create a new item
- In-line "action" buttons controllable on a per-option basis
- In-line editing functionality, with the ability for the user to edit the label of an option and then commit/abandon the edit.

The PR merges this functionality behind a feature flag that will be removed in the future closer to when the course catalog for Spring 2022 is released (fully launching the feature).

### Motivation

#11 is one of the most commonly-requested features for the app (see #41), so implementing it should be very beneficial to our users.

### Contributors

- @kw143 
- @jazevedo620

### DIFF
--- a/src/components/Header/index.tsx
+++ b/src/components/Header/index.tsx
@@ -25,8 +25,10 @@ export default function Header({
   onToggleMenu,
   tabs,
 }: HeaderProps): React.ReactElement {
-  const [{ term, oscar, pinnedCrns }, { setTerm }] =
-    useContext(ScheduleContext);
+  const [
+    { term, oscar, pinnedCrns, allVersionNames, currentVersionIndex },
+    { setTerm, setCurrentVersion, addNewVersion, deleteVersion, renameVersion },
+  ] = useContext(ScheduleContext);
   const terms = useContext(TermsContext);
   const captureRef = useRef<HTMLDivElement>(null);
 
@@ -53,6 +55,15 @@ export default function Header({
           terms,
           currentTerm: term,
           onChangeTerm: setTerm,
+        }}
+        versionsState={{
+          type: 'loaded',
+          allVersionNames,
+          currentVersionIndex,
+          setCurrentVersion,
+          addNewVersion,
+          deleteVersion,
+          renameVersion,
         }}
       />
 
--- a/src/components/App/content.tsx
+++ b/src/components/App/content.tsx
@@ -91,6 +91,7 @@ export function AppSkeletonWithLoadingTerms({
         onToggleMenu={openDrawer}
         tabs={NAV_TABS}
         termsState={{ type: 'loading' }}
+        versionsState={{ type: 'loading' }}
       />
       {children}
       <Attribution />
@@ -128,6 +129,7 @@ export function AppSkeletonWithSwitchableTerms({
         onToggleMenu={openDrawer}
         tabs={NAV_TABS}
         termsState={{ type: 'loaded', terms, currentTerm, onChangeTerm }}
+        versionsState={{ type: 'loading' }}
       />
       {children}
       <Attribution />
--- a/src/components/AppDataLoader/index.tsx
+++ b/src/components/AppDataLoader/index.tsx
@@ -7,7 +7,13 @@ import {
   ScheduleContext,
 } from '../../contexts';
 import { Oscar } from '../../data/beans';
-import { Schedule, defaultSchedule, ScheduleVersion } from '../../data/types';
+import useVersionActions from '../../data/hooks/useVersionActions';
+import {
+  Schedule,
+  defaultSchedule,
+  TermScheduleData,
+  ScheduleVersion,
+} from '../../data/types';
 import {
   StageLoadScheduleData,
   StageLoadTerms,
@@ -72,6 +78,8 @@ export default function DataLoader({
                           oscar={oscar}
                           scheduleVersion={scheduleVersion}
                           updateScheduleVersion={updateScheduleVersion}
+                          termScheduleData={termScheduleData}
+                          updateTermScheduleData={updateTermScheduleData}
                         >
                           {children}
                         </ContextProvider>
@@ -101,6 +109,12 @@ type ContextProviderProps = {
       draft: Draft<ScheduleVersion>
     ) => void | Immutable<ScheduleVersion>
   ) => void;
+  termScheduleData: Immutable<TermScheduleData>;
+  updateTermScheduleData: (
+    applyDraft: (
+      draft: Draft<TermScheduleData>
+    ) => void | Immutable<TermScheduleData>
+  ) => void;
   children: React.ReactNode;
 };
 
@@ -117,6 +131,8 @@ function ContextProvider({
   oscar,
   scheduleVersion,
   updateScheduleVersion,
+  termScheduleData,
+  updateTermScheduleData,
   children,
 }: ContextProviderProps): React.ReactElement {
   // Create a `updateSchedule` function
@@ -144,13 +160,51 @@ function ContextProvider({
     [updateSchedule]
   );
 
+  // Derive the list of all version names
+  const allVersionNames = useMemo<string[]>(
+    () => termScheduleData.versions.map(({ name }) => name),
+    [termScheduleData.versions]
+  );
+
+  // Get all version-related actions
+  const { setCurrentVersion, addNewVersion, deleteVersion, renameVersion } =
+    useVersionActions({ updateTermScheduleData });
+
   // Memoize the context value so that it is stable
+  const { currentIndex: currentVersionIndex } = termScheduleData;
   const scheduleContextValue = useMemo<ScheduleContextValue>(
     () => [
-      { term, oscar, ...castDraft(scheduleVersion.schedule) },
-      { setTerm, patchSchedule, updateSchedule },
+      {
+        term,
+        oscar,
+        currentVersionIndex,
+        allVersionNames,
+        ...castDraft(scheduleVersion.schedule),
+      },
+      {
+        setTerm,
+        patchSchedule,
+        updateSchedule,
+        setCurrentVersion,
+        addNewVersion,
+        deleteVersion,
+        renameVersion,
+      },
     ],
-    [term, oscar, scheduleVersion, setTerm, patchSchedule, updateSchedule]
+    [
+      term,
+      oscar,
+      currentVersionIndex,
+      allVersionNames,
+      scheduleVersion.schedule,
+      setTerm,
+      patchSchedule,
+      updateSchedule,
+      setCurrentVersion,
+      addNewVersion,
+      deleteVersion,
+      renameVersion,
+    ]
   );
 
   return (
--- a/src/components/Button/index.tsx
+++ b/src/components/Button/index.tsx
@@ -8,7 +8,7 @@ export type ButtonProps = {
   className?: string;
   disabled?: boolean;
   href?: string;
-  onClick?: () => void;
+  onClick?: (e: React.SyntheticEvent<HTMLDivElement>) => void;
   children?: React.ReactNode;
 };
 
@@ -43,7 +43,7 @@ export default function Button({
       tabIndex={0}
       onKeyDown={(event): void => {
         // Intercept enter presses
-        if (event.key === 'Enter' && onClick != null) onClick();
+        if (event.key === 'Enter' && onClick != null) onClick(event);
       }}
       role="button"
     >
--- a/src/components/CombinationContainer/index.tsx
+++ b/src/components/CombinationContainer/index.tsx
@@ -38,9 +38,9 @@ export default function CombinationContainer(): React.ReactElement {
         onChange={(newSortingOptionIndex): void =>
           patchSchedule({ sortingOptionIndex: newSortingOptionIndex })
         }
-        value={sortingOptionIndex}
+        current={sortingOptionIndex}
         options={oscar.sortingOptions.map((sortingOption, i) => ({
-          value: i,
+          id: i,
           label: sortingOption.label,
         }))}
       />
--- a/src/components/HeaderDisplay/index.tsx
+++ b/src/components/HeaderDisplay/index.tsx
@@ -1,16 +1,36 @@
 import React from 'react';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import { faBars } from '@fortawesome/free-solid-svg-icons';
+import {
+  faBars,
+  faPencilAlt,
+  faTrashAlt,
+} from '@fortawesome/free-solid-svg-icons';
+import swal from '@sweetalert/with-react';
 
 import { getSemesterName } from '../../utils/semesters';
 import { Button, Select, Tab } from '..';
-import { LoadingSelect } from '../Select';
+import { LoadingSelect, SelectAction } from '../Select';
 import Spinner from '../Spinner';
+import { getNextVersionName } from '../../utils/misc';
 import { DESKTOP_BREAKPOINT, LARGE_MOBILE_BREAKPOINT } from '../../constants';
 import useScreenWidth from '../../hooks/useScreenWidth';
 import HeaderActionBar from '../HeaderActionBar';
+import { softError, ErrorWithFields } from '../../log';
+import useFeatureFlag from '../../hooks/useFeatureFlag';
 
 import './stylesheet.scss';
+
+type VersionState =
+  | { type: 'loading' }
+  | {
+      type: 'loaded';
+      currentVersionIndex: number;
+      allVersionNames: readonly string[];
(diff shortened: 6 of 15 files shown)


kw143:
This pull request addresses issue #11, adding the functionality of creating multiple schedules for each term.

Some minor styling issues still need to be addressed:

- The popup for deleting a version currently uses the same styling as welcome popup; its styling does not change based on theme.
- The buttons in the version drop down displays itself automatically instead of only showing when hovered.
- The buttons in the version drop down are not aligned exactly because longer version name is not shortened with ellipses right now.

kw143:
Fixed remaining styling issues. Convert draft pull request to open pull request.

src/components/App/index.tsx
```diff
@@ -0,0 +1,294 @@
...
+          .reverse();
+        setTerms(newTerms);
+      });
+  }, [setTerms]);
+
+  // Set the term to be the first one if it is unset
+  // (once the terms load)
+  useEffect(() => {
+    const t = Cookies.get('term');
+    if (!term && (!t || t === 'undefined')) {
```
jazevedo620, line 183: What is the point of the up-to-date value `t` here, and in general, why were the changes needed? Additionally, a more descriptive variable name would be appropriate.

src/components/App/index.tsx
```diff
@@ -0,0 +1,294 @@
...
+  // Initialize the versionName to Primary
+  useEffect(() => {
+    const v = Cookies.get('version');
+    if (!versionName) {
+      setVersionName(v || 'Primary');
+    }
+  }, [versionName, setVersionName]);
+
+  // Initialize the version lists for each term
+  useEffect(() => {
```
jazevedo620, line 200: How will this handle when a new term is added to the crawler's data? Will the versions data get an additional element added in that case (either explicitly or implicitly?)

src/components/CombinationContainer/index.js
```diff
@@ -32,12 +32,13 @@ export default function CombinationContainer() {
...
-        // eslint-disable-next-line no-shadow
-        onChange={(sortingOptionIndex) => patchTermData({ sortingOptionIndex })}
         value={sortingOptionIndex}
         options={oscar.sortingOptions.map((sortingOption, i) => ({
-          value: i,
-          label: sortingOption.label
+          innerValue: i,
+          innerLabel: sortingOption.label,
+          // eslint-disable-next-line no-shadow
+          onClick: (sortingOptionIndex) =>
```
jazevedo620, line 40: Is there a reason the onChange became an onClick here?

src/components/Header/index.tsx
```diff
@@ -38,93 +33,70 @@ const Header = ({
...
+    'Quaternary',
+    'Quinary',
+    'Senary',
+    'Septenary',
+    'Octonary',
+    'Nonary',
+    'Denary'
+  ];
+
+  const addVersion = () => {
```
jazevedo620, line 63: Some comments explaining the algorithm being used here would be useful

src/components/Header/index.tsx
```diff
@@ -38,93 +33,70 @@ const Header = ({
...
+    { term, versionName, oscar, pinnedCrns },
+    { setTerm, setVersionName }
+  ] = useContext(ScheduleContext);
   const [terms] = useContext(TermsContext);
-  const [theme, setTheme] = useContext(ThemeContext);
-  const captureRef = useRef<HTMLDivElement>(null);
+  const [versionLists, { patchVersionsData }] = useContext(VersionsContext);
+  const versionList = versionLists[term];
+  const [theme] = useContext(ThemeContext);
+  const [versionIndex, setVersionIndex] = useState(
```
jazevedo620, line 47: What is the difference between the versionIndex and position of versionName in the versionLists[term] list?

src/components/Select/index.js
```diff
@@ -1,36 +1,172 @@
...
 import { Button } from '..';
 import './stylesheet.scss';
+import { ThemeContext } from '../../contexts';
 
-export default function Select({ className, value, onChange, options }) {
+export default function Select({ className, value, options }) {
   const [opened, setOpened] = useState(false);
+  const [inputIndex, setInputIndex] = useState('');
+  const [inputting, setInputting] = useState(false);
+  const [versionName, setVersionName] = useState('');
```
jazevedo620, line 19: A lot of these options seem to specifically mention versionName (and versions) despite this being a general-purpose component. It would probably be better suited to be pulled out and become a specialized VersionSelect that includes these parts (in addition to the base Select code).

src/components/Select/index.js
```diff
@@ -1,36 +1,172 @@
...
+                      swal({
+                        buttons: ['Cancel', 'Delete'],
+                        className: `${theme}`,
+                        content: (
+                          <p>
+                            {/* eslint-disable-next-line max-len */}
+                            {/* eslint-disable-next-line react/no-unescaped-entities */}
+                            Are you sure you want to delete "{innerLabel}
+                            {'" '}
+                            schedule?
```
jazevedo620, line 154: Again, this is a very specific reference to the schedule version delete behavior in a general-purpose component

src/components/App/index.tsx
```diff
@@ -0,0 +1,294 @@
...
+    const vs = Cookies.get('versions');
+    if (!vs || vs === '{}') {
+      patchVersionsData(
+        terms.reduce((ac, cur) => ({ ...ac, [cur]: ['Primary', 'New'] }), {})
```
jazevedo620, line 204: 'New' definitely shouldn't be its own element in the stored version list since it should just be an additional option. Removing this may simplify other logic as well.

jazevedo620 (changes requested):
Since this is a very in-depth PR, I have a few requested changes related to details of the algorithm and requests for clarifications/comments so I can better understand the contents.

Additionally, from my basic testing, I wasn't able to replicate the desired behavior that existing schedules in the app get seamlessly transformed to the single "Primary" schedule version by default when users load this version of the app for the first time (for backwards compatibility). Is this behavior currently in the PR, or does it still need to be added?

As an aside, I added a few commits to:
- upgrade TypeScript
- move the header action row to its own component and place it in the nav drawer on small screen sizes (to leave enough room for the version switch select)
- Update useMedia/useScreenWidth hooks and add types to App/index.js->ts

src/components/App/index.tsx
kw143: This "t" variable reads the term value from cookie instead of relying on the "term" state. When app first loads and reaches this useEffect(), the "term" state is always undefined; if you check the current implementation on gt-scheduler.org, no matter which term you were at when you closed the page, you always go back to Fall 2021 when you reopen the app. This is caused by the issue I mentioned, and my code will read from cookie and go back to the same term the user left at. I'll rename "t" to "term_from_cookie" for clarity.

src/components/App/index.tsx
kw143: This function initializes version lists based on the "terms" state, which is set based on get response from the crawler. Therefore, the initialization will add new lists explicitly when user first use the app. However, if the versions is already set, meaning the user already used the app and a new term is added, current implementation will not add a new list. I added new code to account for that; thanks for catching that!

src/components/CombinationContainer/index.js
kw143: Yes, for consistency reason, I changed all select to change its value based on onClick() function. The general <Select /> now uses onClick = {() => onClick(value)} as opposed to the original onClick = {() => onChange(value)}.

src/components/Header/index.tsx
kw143: I added comments to the function. Thanks for pointing out it was hard to understand.

src/components/Header/index.tsx
kw143: There is no difference. "versionIndex" is  the index of "versionName" in its corresponding version list. In header, "versionIndex" is used to perform various operations to version lists, "versionName" is never used directly except getting the right "versionIndex" at the beginning.

src/components/Select/index.js
kw143: This versionName variable is actually whatever the inputting value is. I just referenced it as versionName because I was working with verionName only. It is not tied with the actual versionName context and version cookie. My bad; I am renaming it to inputValue.

src/components/Select/index.js
kw143: Extracted this swal alert out to the delete function passed into the general <Select /> component. Everything in <Select /> should be general and decoupled from version select. Each option in <Select /> takes in its identifier(optionId), its text(optionLabel), its onClick, its related icons and their corresponding functions. However, each icon's display (before or after the text, when to display), picture have to be specifed within the <Select />.

src/components/App/index.tsx
kw143: I am currently using version lists stored in the cookie as direct inputs into the <Select /> as options. "New" is one of the option and it has an unique icon and onClick passed in as well. We can separate out the "New" from version lists in cookies, although I do not think it will make other implementation any simpler since you would have to manually add the option "New" to the version lists before passing them in as options to <Select /> any way. I do not think it is worth to make this change since it will require changes to many parts and it really does not make other implementation much simpler.

jazevedo620: I just think that storing "New" as an explicit cookie adds no value (at the very least, you could simply append it to the end whenever you need the full array like `[...arr, "New"]`) and makes the state more brittle. Storing it creates an implicit invariant that it must always exist once at the end of the list for each term, and that invariant could get broken unintentionally in the future (perhaps by future contributors/buggy code). If you think it is justified for the sake of simplicity, I suppose it doesn't hurt and we can leave it in, but it does seem like an anti-pattern (at least to me).

src/components/App/index.tsx
```diff
@@ -0,0 +1,329 @@
...
+
+  // Set the term to be the first one if it is unset
+  // (once the terms load)
+  useEffect(() => {
+    const term_from_cookie = Cookies.get('term');
+    if (!term && (!term_from_cookie || term_from_cookie === 'undefined')) {
+      const [recentTerm] = terms;
+      setTerm(recentTerm);
+    } else if (term_from_cookie && term_from_cookie !== 'undefined') {
+      setTerm(term_from_cookie);
```
jazevedo620, line 188: This part just seems to replicate the existing behavior of the useCookie hook. I think it's fine to leave in, though I strongly think that the code in this component is due for a re-write to have clearer loading stages (though that is out of scope of this PR).

src/components/App/index.tsx
```diff
@@ -0,0 +1,329 @@
...
+        );
+      }
+    }
+  }, [patchVersionsData, terms, versionLists]);
+
+  // Backward compatibility for schedule data before version switch implementation
+  // Copy the old cookies to match with the new cookies' naming format.
+  // Checks for term 202008, 202102, 202105, 202108 at the time of implementation.
+  // Does not remove the old cookies.
+  useEffect(() => {
```
jazevedo620, line 234: This is good, and I appreciate the comment to describe where the magic numbers come from. Just to clarify: this works because "Primary" is _always_ included as a key in the `versionLists`, right (so the `versionLists` cookie itself doesn't have to be modified in this code)?

src/utils.js
```diff
@@ -93,6 +93,25 @@ const getSemesterName = (term) => {
   return `${semester} ${year}`;
 };
 
+const getSemesterMonth = (term) => {
```
jazevedo620, line 96: Why was this function added? I can't seem to find any uses of it.

jazevedo620 (changes requested):

jazevedo620:
I think this PR is almost in a really good spot. All of my concerns were addressed, I just had a couple final questions/clarifications.

In terms of a path forwards, I am going to make some minor CSS modifications to the select and perform some testing locally, and then I think it will be feature complete.

However, when it comes to deploying this feature, I want to be careful in how we do this in case there are additional edge cases we didn't consider (so that we don't break any schedules). I'm going to try to see what the best way to do this is (for example whether we need to do some sort of announcement that stuff *might* break and to tell us if it does), and follow-up soon (within a week).

src/components/Header/index.tsx
jazevedo620: When I first thought it, I was worried that it might be a bit of a code smell that the state has been duplicated, but since you say that the versionName is only used initially I think it's fine.

src/components/App/index.tsx
jazevedo620: That makes sense, thanks

jazevedo620:
I made some minor edits to the styling/interaction behavior of the `<Select />` component; please look over my changes and let me know if I broke anything.

Additionally, while testing, I noticed a few minor edge cases purely with the `<Select />` component and editing/deleting:

1. When renaming an option, if you don't make any edits or re-type the existing name and then commit the edit, then instead of nothing happening (as expected), it displays an error message saying "Name is already in use" despite it being the same item that currently has that name.
2. If you go into edit mode by clicking/tapping an item, and then you delete that item, then the Select component is stuck in edit mode and cannot be closed until another item is designated as the active edited item (and then the edit to that option is committed, which lets the Select component be closed again).
3. If you have 2 options in the Select (not including New) and have the second option active, then deleting the second item will result in "New" becoming the next active option instead of the first option.

The last thing I noticed (and I don't expect to be able to fix this) is that on small screen devices in portrait mode, the buttons on the right side of the options in the Select can go off screen and there is no way to interact with them. With that being said, one simple workaround is to rotate the screen to landscape (assuming a mobile device with a portrait/landscape orientations).

src/utils.js
kw143: When I was renaming `innerValue`, `innerLabel` in `<Select />` to better differentiate option label and identifier, at one point I tried to call onClick for term select using label instead of identifier; I needed to convert label ("Summer 2020" format)  to identifier ("202005" format). I created this helper function. Though in the end I did not need to use it, I left it in the utils.js file, just in case if we need the conversion later.

src/components/App/index.tsx
kw143: I agree. I only did the modification to address the problem where the initial term state is not set based on the term cookie. I believe the useCookie() hook did not set the state based on cookie before this useEffect() is called. That is why the current deployment always defaults to Fall 2021 regardless which term you last left the app at. There is probably a better way to do this, but this solution works for now.

src/components/App/index.tsx
kw143: Yes, the initialization of `versionLists` adds `Primary` to the version list of each term. When user switch to the new version, there will be no cookie storing verison lists; the initialization automatically adds in `Primary`'s to the cookie.

src/components/App/index.tsx
kw143: I agree. It would probably be counter-intuitive for other people who might be using version lists in the future. I'll try to take 'New' out of the version lists, and append it like you mentioned using spreading. The reason I did not want to modify this was there are other parts that also need to change besides simply appending 'New' as a option; for example, the indexing of addVersion function. I'll go back and make all the corresponding changes.

jazevedo620:
Hey, @kw143, sorry this feature took so long to get wrapped up. I've been working on this for the past few weeks, first re-writing the core data storage layer in the app (in #69 and #84) to better support adding new features that require data schema changes in a safe manner. With those changes merged into `development` (and running in production), I finally took the time to rebase this PR and clean up some of the logic to better work with the new data storage framework. The original changes are still in their own branch, though I squashed them all into one commit (https://github.com/gt-scheduler/website/tree/keep-me-original-migrate-schedule-version). The result of rebasing is a full implementation of schedule version switching with the ability to add, edit, and delete schedules using the UI you developed.

However, the usage of gt-scheduler varies greatly over time, and right now its metrics show that it has a low active current user count (since it's after phase 2 registration for Fall 2021, and before the catalog has been released for Spring 2022). As a result, I think we should wait to "release" this feature until closer to when the catalog for Spring 2022 is released, at which point we can make the `<InformationModal>` display a new message announcing it release (I might also make a Reddit post on r/gatech).

To facilitate a later release without keeping this PR from being merged, I made it so that the schedule switching UI is hidden behind a feature flag (but not the data storage code; that will still be running) that we can simply disable when it comes time to releasing the feature. All of the code will still be shipped to production, and users/developers can, if they want, manually enable it by running the following command into their devtools console:
```js
window.localStorage.setItem('ff-2021-09-14-schedule-versions', 'true')
```
Once this command is run and the page is refreshed, the schedule switching interface will then be visible (and the opt-in of the feature flag will be persisted to local storage).

With that all in mind, I'm going to add a bit more info to the PR description and merge it :tada:. Thanks for all the hard work you put into this feature :shipit:

jazevedo620 (approved):
:shipit:

PR #26, opened 2021-03-21
Zhen/ 19 Fix am/pm suffix bug

Jz1116:


jazevedo620 (approved):
I made the logic slightly easier to follow just so that it can be more easily changed/understood in the future. Thanks for fixing the underlying bug! :shipit:

PR #30, opened 2021-03-29
Feature/overlapping blocks

JustinH11235:
#### Reason for Feature
Personally, when I make my schedule I tend to pin every section time possible onto my calendar so that I can visually see my options, then I slowly whittle down my selections until I get only one section per course. When switching to gt-scheduler this semester from courseoff, I noticed this approach is difficult when times partially overlap and impossible when they fully overlap.

#### Changes Implemented
1. Within every day, whenever courses time blocks overlap, I modify their 'width' and 'left' CSS properties to make each of these blocks less wide and horizontally separated.
2. Unlike courseoff, I do this in a way so that you can have a grouping of overlapping blocks at the top of the day, each with a width of say 1/3, and have a disjoint grouping of overlapping blocks at the bottom of the day, each with a width of say 1/2.
3. JustinH11235@7fab9359b75a790bceeae17019d3db492b9e0d39 sorts the pinned/overlayCrns from shortest to tallest as a heuristic that attempts to fit the blocks together in a way that minimizes the number of blocks in a row (not necessary but will show before/after pictures below).
## Before PR
schedulerold1

## After PR (without Heuristic) (same courses as before)
Circled in red are cases where we fit the blocks together inefficiently, causing each block's width to be smaller.
schedulernew3

## After PR (with Heuristic)
schedulernew1

I believe this is a necessary feature for many users, but if you have any critiques, additions, changes, etc. please let me know! Also, I am new to open-source contributing so please let me know if I'm doing anything wrong and I'll try to fix it. Hope this is helpful!

### DIFF
--- a/src/components/Calendar/index.js
+++ b/src/components/Calendar/index.js
@@ -23,60 +23,73 @@ export default function Calendar({
   const meetingLen = (m) => m.period.end - m.period.start;
   crns.sort(
     (a, b) =>
-      Math.max(...oscar.findSection(a).meetings.map(meetingLen)) -
-      Math.max(...oscar.findSection(b).meetings.map(meetingLen))
+      Math.max(
+        ...oscar
+          .findSection(a)
+          .meetings.filter((m) => m.period)
+          .map(meetingLen)
+      ) -
+      Math.max(
+        ...oscar
+          .findSection(b)
+          .meetings.filter((m) => m.period)
+          .map(meetingLen)
+      )
   );
 
   crns.forEach((crn) => {
-    oscar.findSection(crn).meetings.forEach((meeting) => {
-      meeting.days.forEach((day) => {
-        let curRowSize = 1;
+    oscar
+      .findSection(crn)
+      .meetings.filter((m) => m.period)
+      .forEach((meeting) => {
+        meeting.days.forEach((day) => {
+          let curRowSize = 1;
 
-        Object.values(dayMap[day])
-          .filter(
-            (entry) =>
-              entry.period.start < meeting.period.end &&
-              entry.period.end > meeting.period.start
-          )
-          .forEach((entry) => {
-            curRowSize = Math.max(curRowSize, entry.rowSize + 1);
-          });
-
-        const updatePrevious = (arr, seen, curCrn, curPeriod) => {
-          if (seen.has(curCrn)) {
-            return;
-          }
-          seen.add(curCrn);
-
-          arr
+          Object.values(dayMap[day])
             .filter(
               (entry) =>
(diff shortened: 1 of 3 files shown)


jazevedo620:
First of all, thanks so much for submitting this pull request 🎉. I really appreciate how detailed the PR is and the extra effort you went through to make it visually appealing.

I will go ahead and take a look at the changes now and get back back to you momentarily.

src/components/Calendar/index.js
```diff
@@ -12,7 +12,81 @@ export default function Calendar({
...
+        meeting.days.forEach((day) => {
+          const curRowSize = Object.values(dayMap[day])
+            .filter(
+              (entry) =>
+                entry.period.start < meeting.period.end &&
+                entry.period.end > meeting.period.start
+            )
+            .reduce((acc, entry) => Math.max(acc, entry.rowSize + 1), 1);
+
+          const updatePrevious = (arr, seen, curCrn, curPeriod) => {
```
jazevedo620, line 54: Can you add some comments explaining the purpose of this function?

src/components/Calendar/index.js
```diff
@@ -12,7 +12,81 @@ export default function Calendar({
   capture,
   isAutosized
 }) {
-  const [{ pinnedCrns }] = useContext(TermContext);
+  const [{ pinnedCrns, oscar }] = useContext(TermContext);
+
+  const dayMap = DAYS.reduce((acc, day) => {
```
jazevedo620, line 17: It would be useful to include a comment about what exactly this data structure contains

src/components/Calendar/index.js
```diff
@@ -12,7 +12,81 @@ export default function Calendar({
...
+      ) -
+      Math.max(
+        ...oscar
+          .findSection(b)
+          .meetings.filter((m) => m.period)
+          .map(meetingLen)
+      )
+  );
+
+  crns.forEach((crn) => {
```
jazevedo620, line 40: Can you add some comments explaining the overview of the algorithm used here? It seems to be both iterative and recursive and use a visited set, but the exact mechanism is unclear.

jazevedo620 (changes requested):
The code all looks good, though some of it is a little unclear on first read. Some brief comments (on the order of 2-5 lines) would be really helpful to document the approach that the algorithm uses and what exactly it outputs.

JustinH11235:
Hopefully the code is clearer after the last commit. While I was cleaning up the code, I decided I didn't like how I was passing more data to TimeBlocks than they needed so I restructured dayMap (renamed to crnSizeInfo) to easily pass only the necessary info.

PR #48, opened 2021-08-23
Fix overlapping map markers

### LINKED ISSUE #47: Markers for classes in the same location overlap
When multiple classes on the same day have the same building location, their markers will overlap and only one of them will show.

Reproduction:
1. Add MATH 4022 AU and MATH 4221 AU to the schedule (Both are located on Skiles on Tuesday)
2. Navigate to Map

image

huangvincent170:
Resolves #47


src/components/MapView/index.js
```diff
@@ -14,8 +14,33 @@ const MapView = ({ locations }) => {
...
+    if (!location.coords) {
+      unknown.push(location);
+    } else if (coordsToLocationMap.has(location.coords)) {
+      coordsToLocationMap.get(location.coords).push(location);
+    } else {
+      coordsToLocationMap.set(location.coords, [location]);
+    }
+  });
+
+  coordsToLocationMap.forEach((coordLocations, coords) => {
```
jazevedo620, line 31: [nit] It would be more natural to use map here; i.e:

```js
const markers = coordsToLocationMap.entries().map(([coordLocations, cords]) =>
  // ...
);
```

src/components/MapView/index.js
```diff
@@ -14,8 +14,33 @@ const MapView = ({ locations }) => {
...
+    markers.push(
+      <Marker latitude={coords.lat} longitude={coords.long}>
```
jazevedo620, line 33: `<Marker>` here should have a `key` prop: https://reactjs.org/docs/lists-and-keys.html (I'd recommend just using the index in the list since these items won't change at runtime)

src/components/MapView/index.js
```diff
@@ -14,8 +14,33 @@ const MapView = ({ locations }) => {
...
+        <FontAwesomeIcon icon={faMapPin} className="pin-icon" />
+        <p className="pin-text">
+          {coordLocations.map((location) => (
+            <div>
```
jazevedo620, line 37: This `div` also needs a `key` prop (maybe the section ID:

```js
`${location.id} ${location.section}`
```

jazevedo620 (changes requested):
I left a couple nits and requests; feel free to update the PR and re-request a review whenever. Also thanks for contributing :heart:

PR #50, opened 2021-08-24
Migrate classes in beans/*.js, utils.js to TypeScript

jazevedo620:
### Summary

This PR includes renames of each file in `beans/` from `*.js` to `*.ts` as well as all needed class field/function types to ensure that the classes are well-typed. Beyond that, there are some minor runtime changes (mostly centered around stricter undefined/null-checking), but *these shouldn't substantially affect the runtime behavior of the app*.

Additionally, many types from the crawler (from [gt-scheduler/crawler/src/types.ts](https://github.com/gt-scheduler/crawler/blob/master/src/types.ts)) were imported to this repo to power the expected types that are used to instantiate the various beans from the crawler data.

### Testing

- Manually running `npm run start` shows the app as normal being functional
- Linting/formatting passes

### Rollout plan

These changes should be safe to rollout immediately since they aren't supposed to change any substantive code paths.

The PR is safe to rollback; we can monitor the issues page, Sentry, and Reddit over the next week or so to see if any issues are reported that might be related.


src/beans/Course.ts
```diff
@@ -20,7 +62,7 @@ class Course {
     const onlyLabs = this.sections.filter(
       (section) => isLab(section) && !isLecture(section)
     );
-    this.hasLab = onlyLectures.length && onlyLabs.length;
+    this.hasLab = !!onlyLectures.length && !!onlyLabs.length;
```
jazevedo620, line 65: This is a substantive change (type is changed from truthy/falsy number to proper boolean), but the only uses of `hasLab` are in conditional statements throughout `beans/` so it should be safe.

src/beans/Feedback.ts
```diff
@@ -1,10 +1,15 @@
...
 
-// Returns a axios chain
-const FormSubmit = ({ rating, feedback }) => {
+const FormSubmit = async ({
+  rating,
+  feedback
+}: {
+  rating: number;
+  feedback: string;
+}): Promise<void> => {
```
jazevedo620, line 10: Changing it to return a void Promise is a substantive change, but the only usage of this (in [src/components/Feedback/index.js](https://github.com/gt-scheduler/website/blob/development/src/components/Feedback/index.js#L20)) doesn't use the resolved value that axios gives, so it shouldn't change anything.

src/beans/Section.ts
```diff
@@ -41,7 +101,7 @@ class Section {
         dateRangeIndex
       ]) => ({
         period: oscar.periods[periodIndex],
-        days: days === '&nbsp;' ? [] : [...days],
+        days: days === '&nbsp;' ? [] : days.split(''),
```
jazevedo620, line 104: This is because TypeScript doesn't like treating the string as an iterable.

src/beans/Section.ts
```diff
@@ -51,14 +111,18 @@ class Section {
       })
     );
     this.instructors = unique(
-      this.meetings.reduce(
-        (instructors, meeting) => [...instructors, ...meeting.instructors],
-        []
-      )
+      this.meetings
+        .map<string[]>(({ instructors }) => instructors)
+        .reduce((accum, instructors) => [...accum, ...instructors], [])
```
jazevedo620, line 116: TypeScript doesn't recognize a reduce with different types for the source array & accumulator (as in `Array<A>.reduce<B>((accum: B, elem: A) => B, start: B): B`) unfortunately, so we just move the transformation to an explicit map step and then reduce the `string[][]` with a `string[]` accumulator.

src/utils.ts
```diff
@@ -53,25 +57,27 @@ const hasConflictBetween = (section1, section2) =>
...
-const simplifyName = (name) => {
+const simplifyName = (name: string): string => {
   const tokens = name.split(' ');
   const firstName = tokens.shift();
   const lastName = tokens.pop();
   return [firstName, lastName].join(' ');
 };
 
-const unique = (array) => [...new Set(array)];
+const unique = <T>(array: T[]): T[] => Array.from(new Set(array));
```
jazevedo620, line 72: TypeScript doesn't like spreading the Set into an array on our target ES script version (es5), so we use `Array.from` instead.

src/utils.ts
```diff
@@ -102,24 +108,27 @@ const humanizeArray = (array, conjunction = 'and') => {
...
-    reqs.slice(1, reqs.length).forEach((req, i) => {
+  } else if (reqs[0] === 'or') {
+    const [, ...subClauses] = reqs;
+    subClauses.forEach((req, i) => {
       string += decryptReqs(req) + (last(i) ? '' : ' or ');
     });
-  else
-    reqs.forEach((req, i) => {
-      string += req.id + (i === reqs.length - 1 ? '' : ' or ');
-    });
```
jazevedo620, line 122: I'm not sure why this branch was there in the original code... it shouldn't be possible to hit from what I can tell.

Issue #27, opened 2021-03-22
Plans for an account feature?


tumbleshack:
Love your work here! Absolutely incredible; the best scheduler tool to use, no question.

For my friends and I, however, the essential feature for this scheduler to take over from CourseOff is the ability to save my schedule with my account, so I can access it on any device and not lose it every time that I clear cookies.

Are there plans for this? If so, do you need any help?

Anshuljkt:
this!

karpawich:
@tumbleshack one day! We hope so. I'll let the rest of the team speak on this (although they may be busy), but I can tell you that Bits of Good will be very careful in adding any sort of backend component that stores student data. In addition to security requirements, we also will need to engage in best practices for ethical data access and management. It is completely within the realm of possibility, but it will take time to make sure that we do it right!

(Max - Exec Director)

tumbleshack:
Great. So I take it that community contributions to this feature are not welcome?

abhaycashikar:
Another way I've seen applications handle this problem instead of using accounts is using some kind of unique code (like a hash) that represents the user's application state. A user could export a code from one device and enter that same code on the other device to automatically add all the same classes. Something like that wouldn't require any storage of student data, which is the major issue with any account feature IMO.

jazevedo620:
Thanks for the feature request! I spoke with some of the Bits of Good team that co-maintains this project and this is definitely a highly requested feature that is on our radar. In terms of whether we have the bandwidth as a team to finish it this semester, it mostly comes down to how and when Bits of Good works on projects. We have about 3 more weeks of active development before we recess for finals/summer break, and in those 3 weeks of time, we have a set of tasks that we're looking to finish implementing this semester (namely, schedule version switching, bugfixing, and a course catalog feature). With that in mind, this is definitely a feature that will be prioritized in the future, due to the demand and usefulness that it brings, even if that means it might have to wait until fall of this year to be actively worked on.

> Great. So I take it that community contributions to this feature are not welcome?

[...]

PR #84, opened 2021-09-12
Migrate to local storage, add data-layer support for schedule versions

jazevedo620:
### Summary

This PR includes some major changes to the data layer of the application, many of which build on the work done in #69. Specifically, this PR contains a major migration of the backing storage of user data from cookies to local storage (see Motivation for the reasoning). Local storage is now used to store:
- Core schedule data
- UI theme
- [`<InformationModal>`](https://github.com/gt-scheduler/website/blob/development/src/components/InformationModal/index.tsx) seen flag

Along with the migration to local storage, I created a more reliable framework for implementing, running, and testing **migrations** on the schema of the stored schedule data (borrowing principles from the more general practice of [schema migrations from the database world](https://en.wikipedia.org/wiki/Schema_migration)). Whether to run a migration is determined by the `version` field, which identifies the schema version for the data. Then, by judiciously using TypeScript, the framework ensures that we migrate data to the latest schema before accessing it in the rest of the application (re-using the work done in #69 to wait to render the app until migration is done). To be more confident in the safety of these changes, I also added a comprehensive test suite for the core hooks/migration logic involved that includes mocks for the cookies and local storage APIs.

Finally, this PR includes a few other smaller but related changes:
- Core schedule data is now stored in a **single data key** (`schedule-data`). To make modifying this state tree easier, I introduced the usage of [Immer](https://immerjs.github.io/immer/) to make callbacks that can update the state using so-called "drafts" (copy-on-write mutable handles that perform immutable updates on the backing data once committed)
- In the first version of the new versioned schema, I also included support for including multiple schedule versions (though nothing in the UI surfaces this capability yet). My plan is to merge this and then rebase #23 to add the existing UI built there on top of the new data-layer capabilities.
- I formalized the pattern introduced in #69 where we have a series of components that render their children conditionally based on asynchronous data fetching state. Now, each of these components are called "stages", and they are orchestrated in the `<AppDataLayer>` component, which handles running all stages before providing the valid context values to their corresponding providers.

#### Moving to local storage from cookies

By migrating from storing user data in cookies to local storage, we gain a few key benefits:
- Effectively remove the cap on the size of the data stored. Cookies have a [maximum size of 4 KiB](https://docs.microsoft.com/en-us/previous-versions/ms178194(v=vs.140)?redirectedfrom=MSDN#cookie-limitations), whereas local storage can be at least 5 MB (a limit that schedule data will almost certainly never hit).
- Because the size limitation is now removed, this also opens up the opportunity to consolidate all schedule data into a single state tree, making update logic simpler and opening up the opportunity to more easily sync the schedule data in the future (#27)
- Cookies are also limited in *how many of them* we can have per-site, [with most browsers implementing a limit around 20](https://docs.microsoft.com/en-us/previous-versions/ms178194(v=vs.140)?redirectedfrom=MSDN#cookie-limitations). The scheduler was already getting dangerously close to this limit (for example, I have 15 cookies for gt-scheduler.org on my own computer), and the addition of schedule versions would likely cause it to start hitting the limit.
- Using cookies adds a small slow-down to the site, since the cookies are tacked on to every request the browser makes to the gt-scheduler.org domain. Unfortunately, this PR doesn't actually remove the existing cookies, but it will stop new users from incurring the slow-down.
  - The plan is to roll out these changes, and then in a patch a few months down the line (once registration for the next term starts and this new storage framework is battle tested), remove all existing data-related cookies (to remove their overhead).

#### Introduction of migration framework

The motivation behind the migrations is ultimately to:
- increase our confidence in the validity of the data stored in users' browsers (since we don't have access to it),
- to make it much easier to confidently change the schema in the future to support new features. The combination of TypeScript and testing lets us greatly increase our confidence in any future changes, and my hope is that it makes for a stronger codebase.

#### Adding data-layer support for schedule versions

These changes make it much easier to merge #23, which has unfortunately gone unmerged for too long (which is entirely my fault).


.github/workflows/ci.yml
```diff
@@ -49,6 +49,6 @@ jobs:
         run: yarn install --frozen-lockfile
 
       - name: Test
-        run: yarn run test
+        run: yarn run test --coverage
```
jazevedo620, line 52: TODO remove this change

jazevedo620: Done

PR #134, opened 2022-08-28
Implement Finals Schedule

### LINKED ISSUE #92: Add finals schedule for courses
**The feature**
Give a clear 2-3 sentence description of your feature here.

**Detailed explanation**
Elaborate on your initial description of the feature, including its components.

**Mockups**
If available, include design mockups or examples of similar features in existing products that others can use as reference.

**Additional context**

adityavidyadharan:
Resolves #92 

Creates new tab with finals schedule using information scraped from the Registrar's PDF

Relies on gt-scheduler/crawler#12

Screen Shot 2022-08-28 at 2 34 00 AM

### DIFF
--- a/src/components/Finals/index.tsx
+++ b/src/components/Finals/index.tsx
@@ -1,0 +1,157 @@
+import React, { useContext } from 'react';
+
+import { CLOSE, OPEN } from '../../constants';
+import { timeToShortString } from '../../utils/misc';
+import { ScheduleContext } from '../../contexts';
+import { FinalBlocks } from '..';
+import { makeSizeInfoKey, FinalBlockPosition } from '../FinalBlocks';
+import { Period } from '../../types';
+
+import './stylesheet.scss';
+
+export default function Finals(): React.ReactElement {
+  const [{ pinnedCrns, oscar }] = useContext(ScheduleContext);
+
+  // Contains the rowIndex's and rowSize's passed into each crn's TimeBlocks
+  // e.g. crnSizeInfo[crn][day]["period.start-period.end"].rowIndex
+  const crnSizeInfo: Record<
+    string,
+    Record<string, Record<string, FinalBlockPosition>>
+  > = {};
+
+  // Recursively sets the rowSize of all time blocks within the current
+  // connected grouping of blocks to the current block's rowSize
+  const updateJoinedRowSizes = (
+    periodInfos: FinalBlockPosition[],
+    seen: Set<string>,
+    curCrn: string,
+    curPeriod: Period,
+    newRowSize: number
+  ): void => {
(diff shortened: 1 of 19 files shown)


adityavidyadharan:
The scraper links need to be changed back to point to the gt-scheduler github rather than my account which I had done for testing [here](https://github.com/gt-scheduler/website/pull/134/files#diff-80752a7de78e886041a501178ca7480490cfadeb2e433489b1a7ee6a4cb1588fR11) and [here](https://github.com/gt-scheduler/website/pull/134/files#diff-41aacc606d61cc6e780db4660734f67e7a2e953db5dad082a5edc06f5a6546ccR15)

.github/workflows/ci.yaml
```diff
@@ -10,7 +10,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Install
-        run: yarn install --frozen-lockfile
+        run: yarn install --frozen-lockfile --ignore-engines
```
jazeved0, line 13: I don't think the changes to this file are necessary now that I merged #136. Sorry about that.

jazeved0 (changes requested):
Looks mostly good to me, I just had a few small changes. Thanks a ton for taking the time to implement this!

src/data/beans/Section.ts
```diff
@@ -116,6 +118,14 @@ export default class Section {
           from: new Date(),
           to: new Date(),
         },
+        finalDate:
+          finalDateIndex === -1 || oscar.version < 3
+            ? new Date()
+            : oscar.finalDates[finalDateIndex] ?? new Date(),
```
jazeved0, line 124: Any objections to using `null` instead of `new Date()` here? I think it makes the error/non-existant case much clearer.

src/components/Finals/index.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+  if (oscar.finalDates.length === 0)
+    return (
+      <div className="FinalsEmptyContainer">
+        <h2 className="FinalsEmpty">
+          Finals Schedule not available for this semester
+        </h2>
+        <p className="FinalsEmpty">
+          Find out more about updating it{' '}
+          <a href="https://github.com/adityavidyadharan/crawler/tree/aditya/92-finals-schedule#:~:text=be%20found%20here-,Updating,-the%20list%20of">
```
jazeved0, line 133: For tracking: this also needs to be updated once the crawler PR is merged

src/data/beans/Section.ts
adityavidyadharan: Yeah that makes sense. Not sure why I was creating a new Date object in the first place

jazeved0 (approved):

PR #135, opened 2022-08-30
Clone Schedule Functionality

### LINKED ISSUE #128: Have way to copy schedule between schedule versions
Having some way to copy a schedule version in its entirety or copy some courses to another schedule version would be nice. Alternately, a duplicate schedule version button might also solve this.

adityavidyadharan:
Resolves #128 

Adds a button to clone a schedule in its current state

Screen Shot 2022-08-30 at 1 10 00 AM

### DIFF
--- a/src/components/HeaderDisplay/index.tsx
+++ b/src/components/HeaderDisplay/index.tsx
@@ -209,12 +209,7 @@ function VersionSelector({ state }: VersionSelectorProps): React.ReactElement {
             type: 'button',
             icon: faCopy,
             onClick: () => {
-              state.cloneVersion(
-                version.id,
-                getNextVersionName(
-                  state.allVersionNames.map(({ name }) => name)
-                )
-              );
+              state.cloneVersion(version.id, `Copy of ${version.name}`);
             },
           });
 
--- a/src/components/HeaderDisplay/index.tsx
+++ b/src/components/HeaderDisplay/index.tsx
@@ -4,6 +4,7 @@ import {
   faBars,
   faPencilAlt,
   faTrashAlt,
+  faCopy,
 } from '@fortawesome/free-solid-svg-icons';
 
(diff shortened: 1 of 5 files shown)


src/components/HeaderDisplay/index.tsx
```diff
@@ -202,6 +204,20 @@ function VersionSelector({ state }: VersionSelectorProps): React.ReactElement {
...
+          // Add the duplicate button
+          actions.push({
+            type: 'button',
+            icon: faCopy,
+            onClick: () => {
+              state.cloneVersion(
+                version.id,
+                getNextVersionName(
+                  state.allVersionNames.map(({ name }) => name)
+                )
```
jazeved0, line 216: Thoughts on using this instead? I think it makes it clearer what the new schedule is, though feel free to disagree.

```suggestion
                `Copy of ${version.name}`
```

src/data/hooks/useVersionActions.ts
```diff
@@ -147,5 +149,36 @@ export default function useVersionActions({
     [updateTermScheduleData]
   );
 
-  return { addNewVersion, deleteVersion, renameVersion };
+  const cloneVersion = useCallback(
```
jazeved0, line 152: 👍 nice clean implementation.

Initially, I was slightly concerned that having two calls to `updateTermScheduleData` (one inside `addNewVersion`, and one here) might cause two separate renders if they aren't batched, but I just checked it locally and it only re-renders once in the React DevTools Profiler. No problems here 👍

jazeved0 (changes requested):
Very minor suggested changes, overall great 👍

src/components/HeaderDisplay/index.tsx
adityavidyadharan: Sounds good

jazeved0 (approved):
Looks good to me! :shipit:

PR #142, opened 2023-01-14
Use non-default uid for ical export

### LINKED ISSUE #141: ical export event UID conflict
**Describe the bug**
Export to `ical` uses event UIDs which conflict on the `default` UID domain. Current behavior is to assign UIDs `0, 1, 2, ...` in the `default` domain. This causes any other events in the `default` domain with the same UIDs will be overwritten upon `ical` import, for example, courses imported last semester.

**To Reproduce**
1. Create a schedule with classes from most recent semester
2. Export calendar file
3. Import calendar file in Google calendar
4. Create a second schedule with classes from previous semester
5. Export calendar file
6. Import calendar file in Google calendar
7. Classes from most recent semester will no longer all be present as there is an event UID conflict

**Expected behavior**
UIDs should not be conflicting, schedule from different terms should not replace each other in a calendar environment.

**Additional context**
Consider this generated event for example, in particular the `UIDL6@default` line.
```
BEGIN:VEVENT
UID:6@default
CLASS:PUBLIC
DESCRIPTION:VIP Proj Team: SR I
RRULE:FREQ=WEEKLY;UNTIL=20221216T000000Z;BYDAY=FR
DTSTAMP;VALUE=DATE-TIME:20230114T000546
DTSTART;VALUE=DATE-TIME:20220826T153000
DTEND;VALUE=DATE-TIME:20220826T162000
LOCATION:TBA
SUMMARY;LANGUAGE=en-us:VIP 4601
TRANSP:TRANSPARENT
END:VEVENT
```

kieranlblack:
The events exported are now assigned a domain of `gt-scheduler` along with a random UID. Random UID generation ensures a scenario such as friends importing each other's schedules for the same semester has well-defined behaviour.

An example event can be seen here:
```
BEGIN:VEVENT
UID:03C71D76-9736-4885-92BC-2B81EDC55E8A@gt-scheduler
CLASS:PUBLIC
DESCRIPTION:Abstract Algebra II
RRULE:FREQ=WEEKLY;UNTIL=20230505T000000Z;BYDAY=MO,WE,FR
DTSTAMP;VALUE=DATE-TIME:20230114T012702
DTSTART;VALUE=DATE-TIME:20230109T123000
DTEND;VALUE=DATE-TIME:20230109T132000
LOCATION:Skiles 169
SUMMARY;LANGUAGE=en-us:MATH 4108
TRANSP:TRANSPARENT
END:VEVENT
```

closes #141

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,8 @@
     "react-virtualized": "^9.21.2",
     "s-ago": "^2.2.0",
     "tslib": "2.3.0",
-    "use-local-storage-state": "^11.0.0"
+    "use-local-storage-state": "^11.0.0",
+    "uuid": "^9.0.0"
   },
   "devDependencies": {
     "@bitwarden/cli": "^2022.8.0",
--- a/src/hooks/useHeaderActionBarProps.ts
+++ b/src/hooks/useHeaderActionBarProps.ts
@@ -26,13 +26,13 @@ export type HookResult = Pick<
 export default function useHeaderActionBarProps(
   captureRef: React.RefObject<HTMLDivElement>
 ): HookResult {
-  const [{ oscar, pinnedCrns }] = useContext(ScheduleContext);
+  const [{ term, oscar, pinnedCrns }] = useContext(ScheduleContext);
   const [theme] = useContext(ThemeContext);
   const accountState = useContext(AccountContext);
 
   const handleExport = useCallback(() => {
     try {
-      exportCoursesToCalendar(oscar, pinnedCrns);
+      exportCoursesToCalendar(term, oscar, pinnedCrns);
     } catch (err) {
       softError(
         new ErrorWithFields({
@@ -44,7 +44,7 @@ export default function useHeaderActionBarProps(
         })
       );
     }
-  }, [oscar, pinnedCrns]);
+  }, [term, oscar, pinnedCrns]);
 
   const handleDownload = useCallback(() => {
     const captureElement = captureRef.current;
--- a/src/hooks/useHeaderActionBarProps.ts
+++ b/src/hooks/useHeaderActionBarProps.ts
@@ -26,13 +26,13 @@ export type HookResult = Pick<
 export default function useHeaderActionBarProps(
   captureRef: React.RefObject<HTMLDivElement>
 ): HookResult {
-  const [{ term, oscar, pinnedCrns }] = useContext(ScheduleContext);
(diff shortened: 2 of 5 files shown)


jazeved0 (commented):
The code changes look good to me, and I appreciate the contribution.

My only thought is whether the behavior could be improved even further from the (admittedly pretty bad) current behavior of using basic integer UIDs?

Specifically, do you think it would be a useful feature for a specific course to replace itself if the same or similar schedule was re-imported into a Calendar program (like Gcal)? For example, imagine that I have a course schedule that contains two courses: "A" and "B" which each have a single section and that section has a single "meeting". Then, I export the schedule and add it to Gcal. Then, I decide that I want to add course section "C" to my schedule, so I re-export the schedule and re-import it into Gcal. Should the newly-imported "A" and "B" "events" (in the ICS parlance) override the already-imported events? If I understand correctly (and based on some basic testing I did), the PR currently won't have this behavior, and instead the user would see a duplicate "A" and "B" section in Gcal.

image

I believe that it would be possible to accomplish this by generating the UID deterministically from the tuple of (term, CRN, meeting index), since those together should be globally unique. Something like `UID:course-${term}-${crn}-${meetingIndex}@gt-scheduler`.

What are your thoughts? Do you think this would be a useful behavior (since I'm guessing you're a user of the calendar export)? If you don't think it would be useful, I'm happy to approve the PR as-is.

kieranlblack:
@jazeved0 this is a good point. My thoughts are that I would personally rather leave the functionality as is in the PR right now for two main reasons:
1. We want to provide functionality where users can import their friends' schedules. I think it makes sense for shared classes to be treated as different events, in order to do this deterministically we would need to encode user information in the UID. We can't do this because a user won't necessarily be logged in and also we probably shouldn't put information which directly identifies the user in the UID.
2. I think from a UX perspective, deleting events in their calendar without warning might be confusing. E.g. if a user had made special notes in the event in their calendar and then on reimport the notes disappeared.

I guess in theory we should really have some mechanism for guaranteeing the UIDs are globally unique, but hopefully the 128 bits of randomness are good enough™️.

jazeved0 (approved):
Thanks for the thoughtful response. Those are good points, so I'm happy to approve the PR as-is.

The randomness of the UIDs will be more than sufficient to prevent collisions; there's a reason they're called *universally* unique IDs after all :P (especially since `uuid` uses a cryptographically secure PRNG so it should be well-behaved).

PR #144, opened 2023-01-18
Add "focusing" meetings on calendar, upgrade react-tooltip to v5

### LINKED ISSUE #132: Cannot read location with hover
**The feature**
Give a clear 2-3 sentence description of your feature here.
Would like to see the full location, including room number when hovered over. 
**Detailed explanation**
Elaborate on your initial description of the feature, including its components.
Now I've to manually expand the width of my browser to see the full location 
**Mockups**
If available, include design mockups or examples of similar features in existing products that others can use as reference.
This is the current view: 
image

Expanded view: 
image

Expected view: 
image

**Additional context**

### LINKED ISSUE #133: Show class location and professor in mobile site
**The feature**
Show the class location and professor on the "Courses" and/or "Calendar" section on the mobile site (specifically any device with a horizontal resolution <1024px). If the text is too long to display fully and exceeds the height of the box, then there should be an option to click on the box to open up some sort of dialog box where we can then see the class location and professor. 

**Detailed explanation**
I'm not able to see the class location and professor in the mobile site. It seems like this information is hidden with `display: none` since having this information displayed can sometimes exceed the height of the box (in which the height is based on the duration of the class). 

Although there is already a map view where we can see the general building of where the class takes place, having this location information displayed would be useful since students often need to know the exact room number and location of the class from their phone while commuting. 

**Additional context**
Here's how it looks on desktop:
Screen Shot 2022-08-23 at 8 50 36 PM

Here's how it looks on mobile:
Screen Shot 2022-08-23 at 8 50 25 PM

jazeved0:
### Summary

This PR implements a way to "focus" meeting blocks on the central calendar view, which shows the "details" popover persistently:

#### Desktop demo

desktop

#### Mobile demo

mobile

This is in addition to the previous behavior of showing the details popover when hovering over a meeting (on desktop). This PR builds on that functionality to:
- allow text to be copied/selected in the popover
- make the mobile experience better, since the courses can be intentionally focused by tapping on/away from them
- make the details on the calendar accessible to keyboard users

#### Upgrading to `react-tooltip` v5

This PR also involved upgrading to `react-tooltip` v5, which contains advanced features used in the implementation of the PR. This was made possible now that the project is on React v18 (after #143).

### Motivation

Fixes #133 by allowing the location and professor to be viewed on mobile via the details popover.

Also fixes #132

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "react-overlays": "^5.1.1",
     "react-resize-panel": "^0.3.5",
     "react-scripts": "5.0.1",
-    "react-tooltip": "^4.2.10",
+    "react-tooltip": "^5.5.1",
     "react-transition-group": "^4.4.2",
     "react-virtualized": "^9.21.2",
     "s-ago": "^2.2.0",
--- a/src/components/ActionRow/index.tsx
+++ b/src/components/ActionRow/index.tsx
@@ -10,8 +10,7 @@ type FontAwesomeProps = React.ComponentProps<typeof FontAwesomeIcon>;
 export type Action = {
   icon: FontAwesomeProps['icon'];
   styling?: React.CSSProperties;
-  dataTip?: boolean;
-  dataFor?: string;
+  id?: string;
   onMouseEnter?: () => void;
   onMouseLeave?: () => void;
 } & Omit<ButtonProps, 'children'>;
@@ -41,15 +40,7 @@ export default function ActionRow({
             .flatMap((action) => (action != null ? [action] : []))
             .map(
               (
-                {
-                  icon,
-                  styling,
-                  dataTip,
-                  dataFor,
-                  onMouseEnter,
-                  onMouseLeave,
-                  ...rest
-                },
+                { icon, styling, id, onMouseEnter, onMouseLeave, ...rest },
                 i
               ) => (
                 <Button className="action" {...rest} key={i}>
@@ -57,8 +48,7 @@ export default function ActionRow({
                     fixedWidth
                     style={styling}
                     icon={icon}
-                    data-tip={dataTip}
-                    data-for={dataFor}
+                    id={id}
                     onMouseEnter={onMouseEnter}
                     onMouseLeave={onMouseLeave}
                   />
--- a/src/components/App/index.tsx
+++ b/src/components/App/index.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { TooltipProvider } from 'react-tooltip';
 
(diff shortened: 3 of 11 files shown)


atlant04:
This looks great, thanks a lot! I thing I noticed is that the location gets clipped on the class card itself in the calendar view, which also shows up on exported images. Thoughts on making the text not clipped (no ellipses) and just making it flow on the next line?

jazeved0:
It's possible to entirely disable the text clipping behavior and allow it to flow, but I don't think it's an entirely satisfactory solution.

For example, when the location  is long (like anything in the CULC), it's possible for the instructor(s) (which is arguably just as important) to get cut off entirely:

image

Also, if the height isn't sufficient, then it's possible for the course name itself out of bounds too (though this can probably be worked around):

image

I think one of the the best (and easiest) solutions is just to abbreviate the locations for some of the longer builder names, which I implemented in #154:

image

atlant04:
That sounds like a good compromise. Thanks for looking into this!

atlant04:
I wonder if we could include the details in some sort of scroll view and if the details bounds exceed the card frame then the scroll view would activate and let the users scroll the details. That way, you could let it flow and just have a vertical scroll.

jazeved0:
That's definitely possible, but in my opinion isn't an ideal solution because:
1. It's cramming a scroll container into a place one wouldn't expect (small event boxes on a calendar)
   - it might not be obvious that the content even scrolls, since the instructor might be cut off completely but the rest of the content fits into the base viewport:
   
     mockup

2. It won't help with schedule screenshots (although admittedly neither did this PR)
3. There's already a better (in my opinion) way to get that information in its full form: use the tooltip/popover by hovering/selecting the block.

With that said, I don't think it's a bad solution, and it's definitely possible to implement. If you still think it would be a useful feature, feel free to file a new issue to request it. I probably won't have much time to work on this repo much more since I graduated, but the Bits of Good time might be able to look at it.

atlant04:
I agree, it looks a bit awkward and probably hard to use on 50-minute classes. The tooltips do a great job, I wonder if there should be an option to export the jpg of the calendar with tooltips so that all information is visible. Or have like a legend at the bottom of the calendar with more details? 
Thanks a lot for looking into this!

jazeved0:
> I wonder if there should be an option to export the jpg of the calendar with tooltips so that all information is visible. Or have like a legend at the bottom of the calendar with more details? 

I personally think that'd be a great feature for the app to have👍 (especially the idea of augmenting the export with a legend). I'll file an issue to request that feature in case anyone wants to implement that in the future.

PR #155, opened 2023-01-29
Update Course Combination Algorithm

### LINKED ISSUE #151: Update Course Combination Algorithm
### Description
We need to update the algorithm that generates course combinations so that it does not generate combinations that conflict with the recurring events. This function is called `getCombinations` and you can find it under `src/data/beans/Oscar.ts`.

### Acceptance Criteria
- [x] Generated combinations do not conflict with the events.
- [x] Update `getCombinations`'s usage in `CourseContainer`.

### Other Notes

I'd recommend passing a new parameter `events` (type `Event`) into `getCombinations` and use this parameter to exclude combinations that conflict with the events. I don't expect any changes should be made to the return values with this approach, but let me know if you disagree!

samarth52:
### Summary

Resolves #151

### Checklist

- [x] Generated combinations do not conflict with the events.
- [x] Update getCombinations's usage in CourseContainer.

### How to Test

- In `src/components/CombinationContainer/index.tsx`, assign a dummy events list of type `Immutable<Event[]>` to `events` after the `events` prop import from ScheduleContext
- Choose a term in the dev environment and add courses
- Experiment with different Event times by updating `events`, and check for conflicts in the Combinations tab (the combinations where sections conflict with the recurring event should not appear)
- An example `events`, which will reserve the time between 9:30am to 5:00pm on Monday and Wednesday.
```
events = castImmutable([
  {
    id: '123',
    name: 'test',
    period: {
      start: 570,
      end: 1020,
    },
    days: ['M', 'W'],
  },
]);
```


nhatnghiho (approved):
Lgtm 🎉

PR #157, opened 2023-02-02
Create Event component

### LINKED ISSUE #150: Implement a Component to View and Edit Custom Blocks
### Description
We need a way for users to view all their custom blocks/events and make changes as needed. In this ticket, you will implement a card called `Event` that displays information about the events and contains buttons to edit, change color, and delete the event.

<p align="center">image 
<p/>

Refer to [Figma](https://www.figma.com/file/CkystexX4qQO9OnZRxktXa/GT-Scheduler-%2F-Spr21?node-id=3189%3A1465&t=sYxWyAJgq7O9RQla-0) for more design details.

#### UI Requirements
- [x] Event name field can handle text overflow.
- [x] Works with mobile view.

#### Functional Requirements
- [x] Use `ScheduleContext`, which stores info about custom blocks, to populate and color the card.
- [x] Users can change color with the palette icon.
- [x] Delete icon deletes the event from `ScheduleContext`.

### Other Notes
- `Event` should be quite similar to an existing component called `Course`. Refer to `Course` to make your life easier :).
- All new components must be added to `src/components/index.tsx`.

sophiazlin:
### Summary

Resolves #150 

Creates Event component for custom blocks / events.

#### UI Requirements
- [x] Event name field can handle text overflow.
- [x] Works with mobile view.

#### Functional Requirements
- [x] Use `ScheduleContext`, which stores info about custom blocks, to populate and color the card.
- [x] Users can change color with the palette icon.
- [x] Delete icon deletes the event from `ScheduleContext`.

### How to Test
- In `src/components/CourseContainer/index.tsx` create a list of test Events of type `Event[]`
- Add `Event` components in `src/components/CourseContainer/index.tsx` using test events, similar to how `Course` is added

### DIFF
--- a/src/components/Event/index.tsx
+++ b/src/components/Event/index.tsx
@@ -17,7 +17,7 @@ export type CourseProps = {
   eventId: string;
   eventName: string;
   eventPeriod: Period;
-  eventDays: string[];
+  eventDays: readonly string[];
 };
 
 export default function CustomEvent({
--- a/src/components/Event/index.tsx
+++ b/src/components/Event/index.tsx
@@ -1,0 +1,105 @@
+import React, { useCallback, useContext, useEffect, useState } from 'react';
+import {
+  faPencil,
+  faPalette,
+  faTrash,
+} from '@fortawesome/free-solid-svg-icons';
+
+import { classes, getContentClassName, periodToString } from '../../utils/misc';
+import { ActionRow, Palette } from '..';
+import { ScheduleContext } from '../../contexts';
+import { Event, Period } from '../../types';
+
+import './stylesheet.scss';
+
+export type CourseProps = {
+  eventId: string;
+  eventName: string;
+  eventPeriod: Period;
+  eventDays: string[];
+};
+
+export default function CustomEvent({
+  eventId,
+  eventName,
+  eventPeriod,
+  eventDays,
+}: CourseProps): React.ReactElement | null {
+  const [paletteShown, setPaletteShown] = useState<boolean>(false);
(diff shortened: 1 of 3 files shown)


nhatnghiho (dismissed):
Lgtm 🎉

src/components/Event/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { ScheduleContext } from '../../contexts';
+import { Period } from '../../types';
+
+import './stylesheet.scss';
+
+export type CourseProps = {
+  eventId: string;
+  eventName: string;
+  eventPeriod: Period;
+  eventDays: string[];
```
yatharth-b, line 20: I think we should make this readonly, since that is the type passed in my ScheduleContext

yatharth-b, line 20: I think we should make this readonly, since that is the type passed in by ScheduleContext.

nhatnghiho: We actually don't need to. We can spread `eventDays` using the spread operator to remove the `readonly` tag. But no worries about it. I'll change that when I merge.

yatharth-b: Yes, we can. But I think we should maintain the type of eventDays.

nhatnghiho: I understand what you're saying. But having a mix of `readonly` and non-`readonly` types in the props look kinda cumbersome I'd say. If we really want to maintain the type and keep the code clean, then it'd be best to just pass event id as props.

yatharth-b: that sounds like the best option

nhatnghiho (approved):

PR #167, opened 2023-02-17
Update sort algorithm for "most compact"

### LINKED ISSUE #160: Update Sort Combination Algorithm
### Description
GTS has multiple sort options for course combinations, which are `Most Compact`, `Earliest Ending`, and `Latest Beginning`. The sorting algorithm currently does not take events into account, so we need to update it. This function is called `sortCombinations` and you can find it under `src/data/beans/Oscar.ts`.

### Acceptance Criteria
- [ ] Combinations are sorted appropriately for each setting.
- [ ] Related components (if any) are updated if this change affects them (`Calendar` is an example).

### Other Notes
There are two ways to go about this:
1. Pass `events` to `sortCombinations`.
2. Incorporate `events` into `combinations`. This approach means that you will have to update the `Combination` type, change the return values of `getCombinations`, and update `CalendarProps`.

Go with what you deem is best :)

nathangong:
### Summary

Resolves #160 

Updates the "most compact" sorting option to account for custom blocks

### Checklist

- [x] Combinations are sorted appropriately for each setting.
- [x] Related components (if any) are updated if this change affects them (`Calendar` is an example). -- I don't think anything was impacted by this change

### How to Test
Create a few custom blocks, check how the ordering of the combinations changes as a result


nhatnghiho (approved):
Lgtm 🎉

PR #187, opened 2023-03-03
Samarth/172 load friend schedules

samarth52:
### Summary

Resolves #[ticket-number]

### Checklist

- [ ] Copy paste checklist from issue.


yatharth-b (changes requested):
This would be a big change to all stages, but it is recommended to not use Effects to transform data. 
Read more here - https://react.dev/learn/you-might-not-need-an-effect#how-to-remove-unnecessary-effects

yatharth-b:
The code definitely works, but we need to do more investigation to re-evaluate the stage structure. There's definitely some useEffects which don't need to be there in both the friend stages and the schedule stages, but exactly pin pointing the inefficiencies will take more research in my opinion.

yatharth-b (approved):

nhatnghiho:
Addressed in gt-scheduler/firebase-conf/pull/4, but commenting here for more visibility.

For every request we're making to `fetchFriendSchedule`, the cloud function is triggered twice. It's because when we send data with `application/json`, it would trigger a [CORS pre-flight request](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request) that checks to see if the CORS protocol is understood and a server is aware using specific methods and headers. To [bypass this extra middle step](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#examples_of_access_control_scenarios), we need to decorate our request with the header `Content-Type: application/x-www-form-urlencoded`.

PR #194, opened 2023-03-05
Sophia/migration page down

sophiazlin:
### Summary

Resolves n/a

Adds maintenance page

### Checklist

n/a

### How to Test
Add today's date to (date.getDate() === 15 || date.getDate() === 16) in `App`


src/components/App/index.tsx
```diff
@@ -27,48 +28,60 @@ export default function App(): React.ReactElement {
...
-                If any data is still loading,
-                then it displays an "app skeleton" with a spinner.
-                If there was an error while loading
-                then it displays an error screen. */}
-              <AppDataLoader>
-                <AppContent />
-              </AppDataLoader>
-            </AppNavigation>
-            <Feedback />
+          {!websiteDown && (
```
therealsharath, line 41: NIT: Could use a ternary operator here something like `websiteDown ? <content_if_website_down> : <content_if_website_is_not_down`

therealsharath (changes requested):
Code LGTM other than a tiny nit! Wonder why the preview page is failing, seems to run fun locally so I think you should be good to go? 

@nhatnghiho any guesses why the preview does not work?

src/components/App/index.tsx
nhatnghiho: It's on me lol. The workflow for preview is added in `bog-changes-s23` branch (and any branch that checks out from it) only. `main` doesn't have it. I deployed the migration modal the other day to `main` so all preview sites got deleted

nhatnghiho (approved):

PR #196, opened 2023-03-06
Text for Hidden Courses

### LINKED ISSUE #188: Add text to display Saturday + Async Classes
### Description

Online and non-Atlanta sections tend to be TBA (i.e., having no start and end times on Oscar) and are not displayed on the calendar. Saturday classes, in the mean time, break the calendar and should thus be hidden. As a result, we want to add a text description for hidden sections at the bottom of the calendar.

Image

### Acceptance Criteria
- [x] The bottom text is displayed with all TBA and weekend sections. Each section must include the course name + the section code.
- [x] When a user exports the calendar via downloading the image, the text should also display in the exported image.
- [x] The text should be available in appropriate ratio (text should not be cut off)
  - [x] The text should be wrapped (moved to the next line) or scrollable automatically according to screen size
- [x] Mobile view works.

EmilyAL001:
### Summary

Resolves #188 

Adds text at the bottom of the calendar to inform users of courses that will not appear on the calendar

### Checklist

- [x] The bottom text is displayed with all TBA and weekend sections. Each section must include the course name + the section code.
- [x] When a user exports the calendar via downloading the image, the text should also display in the exported image.
- [x] The text should be available in appropriate ratio (text should not be cut off)
- [x] The text should be wrapped (moved to the next line) or scrollable automatically according to screen size
- [x] Mobile view works.

### How to Test

Add a course that is not on the Atlanta campus or has a Saturday meeting time, and the text of the course name and section will appear at the bottom of the calendar. When you 'download image', the text will also appear at the bottom of the calendar.


therealsharath (changes requested):
Bug: The text appears even when I don't have a section added yet and does not disappear when the "hidden" section is removed. 

image

Additionally, it shows up for all sections of a class (even if the added section is seen on the calendar) as long as one section of the class is async.

image

I believe these are edge cases and not desired behavior. A possible fix could be to try tweaking the filtering logic to resolve these issues.

src/components/Calendar/index.tsx
```diff
@@ -233,6 +234,34 @@ export default function Calendar({
...
+  const finalFilteredCourses = filteredCourses.filter((course) => {
+    let match;
+
+    pinnedCrns.forEach((courseCrn) => {
+      course.sections.forEach((section) => {
+        if (courseCrn === section.crn) {
+          if (
+            section.campus !== 'Georgia Tech-Atlanta *' ||
+            section.meetings[0]?.days.includes('S')
+          ) {
```
nhatnghiho, line 253: Campus shouldn't be used as a filter. We want to filter sections whose period is undefined (i.e. "TBA") since only TBA sections (aside from weekend ones) are hidden from view.

nhatnghiho (changes requested):

nhatnghiho:
Saturday sections should also be hidden. Right now, TimeBlocks for those still show up.

nhatnghiho (approved):

PR #212, opened 2023-03-28
Implement Drag and Drop for Custom Events

### LINKED ISSUE #176: Enable Drag and Drop for Event Blocks
### Description
We want to make Custom Blocks more interactive by allowing users to drag and drop the blocks on the calendar from one date and time to another.

image

Refer to [Figma](https://www.figma.com/file/CkystexX4qQO9OnZRxktXa/GT-Scheduler-%2F-Spr21?node-id=3209%3A2691&t=vZm1CKHbZWDh2XX6-4) for more design details.

### Acceptance Criteria
- [x] User can click on the an `EventBlock` and drag it around.
- [x] While the `EventBlock` is being dragged, the block at the original time should be "blurred"
- [x] Wherever the user drop the `EventBlock`, the event's time info should be updated in `ScheduleContext`.
- [x] UI works for both light and dark modes.

yatharth-b:
### Summary

Resolves #176 

Enabled drag and drop functionality for events. Users can change day and time of events with a single day and only the time of events with multiple days by dragging them. 

NOTE: Events can't be selected on desktop screens anymore, but can do further work to make drag and drop work with select functionality as well if required.

### Checklist
- [x] User can click on the an `EventBlock` and drag it around.
- [x] While the `EventBlock` is being dragged, the block at the original time should be "blurred"
- [x] Wherever the user drop the `EventBlock`, the event's time info should be updated in `ScheduleContext`.
- [x] UI works for both light and dark modes.

### How to Test
Add an event and try dragging it.

### DIFF
--- a/src/components/EventBlocks/index.tsx
+++ b/src/components/EventBlocks/index.tsx
@@ -67,9 +67,7 @@ export default function EventBlocks({
   // Start dragging a block. This event handler is passed down to each block
   const handleMouseDown = (
     e: React.MouseEvent,
-    ref: React.RefObject<HTMLDivElement>,
-    meetingIndex: number,
-    meetingDay: string
+    ref: React.RefObject<HTMLDivElement>
   ): void => {
     if (!ref.current) return;
     if (!deviceHasHover) return;
--- a/src/components/EventBlocks/index.tsx
+++ b/src/components/EventBlocks/index.tsx
@@ -202,7 +202,7 @@ export default function EventBlocks({
                 content: periodToString({
                   start: tempStart ?? event.period.start,
                   end:
-                    (tempStart ?? event.period.start) +
+                    (tempStartRef.current ?? event.period.start) +
                     (event.period.end - event.period.start),
                 }),
               },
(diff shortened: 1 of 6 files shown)


nhatnghiho (changes requested):
Awesome job overall! One small thing to change -- can you round the time of the event when dropping to the nearest 5 minutes? It's kinda hard to precisely drop the block on the exact time down to the minutes.

src/components/EventBlocks/index.tsx
```diff
@@ -39,15 +39,128 @@ export default function EventBlocks({
...
   deviceHasHover = true,
-  selectedMeeting,
-  onSelectMeeting,
+  daysRef,
+  timesRef,
 }: EventBlocksProps): React.ReactElement | null {
+  const [tempStart, setTempStart] = useState<number>(event.period.start);
+
+  const tempStartRef = useRef<number>(event.period.start);
+  const tempDaysRef = useRef<string[]>([...event.days]);
```
nhatnghiho, line 48: When a recurring event occurs more than once a week, the blocks are dragged in conjunction with each other and we also cannot drag them to a different day. We should be able to drag the blocks separately.

yatharth-b: I guess we could seperate the events if any single block is moved, but that would cause a lot of accidental seperations

nhatnghiho: Ah, I see what you mean. Then this should be fine. We don't really want to separate events for now.

PR #221, opened 2023-04-05
Recurring Events Export

samarth52:
### Summary

Resolves: N/A

With the addition of recurring events, we want users to be able to export their events as well into a `.ics` calendar or image. However, having only recurring events in the schedule does not enable this feature. Moreso, the exported `.ics` calendar does not contain recurring events. This PR fixes these issues and other minor bugs.

### Checklist

- [x] Fixed export not enabling with only recurring events
- [x] Fixed export to calendar not exporting recurring events
- [x] Fixed not being able to set start/end time between 9 PM and 10 PM
- [ ] Rounded event start and end times to multiples of 5 on submit in the EventAdd tab

### How to Test
- Add events and try exporting to calendar or image

### DIFF
--- a/src/components/EventAdd/index.tsx
+++ b/src/components/EventAdd/index.tsx
@@ -101,8 +101,6 @@ export default function EventAdd({
   );
 
   const onSubmit = useCallback((): void => {
-    const roundedStart = Math.floor(parseTime(start) / 5) * 5;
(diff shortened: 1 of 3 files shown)


nhatnghiho (approved):
Lgtm 🎉

PR #242, opened 2023-06-25
Abhiram/122 import crns

### LINKED ISSUE #122: Import CRNs back into scheduler
The ability to import CRNs and the corresponding class show up on the scheduler.

abhiramg2021:
### Summary

Resolves #122 

Adds the ability to add one class at a time to the schedule using a CRN.

### How to Test
There is now a toggle when you are searching for a class. If you click on the word "ID", it changes to the word "CRN", and you can type in a CRN to add a course to the schedule using CRNS.


abhiramg2021:
I ran yarn build on both the master branch and my feature branch on my machine, and they both built without errors. I'm not sure what seems to be reason for why my code is failing the build in this pipeline.

PR #270, opened 2024-01-26
Toast Component for Undecided Schedules

### LINKED ISSUE #266: [BoG S24] Create Toast Notification for Undecided Schedules
### Problem Description
In Fall 2023, PMT worked on adding crawler support for identifying which semesters' schedules have not been decided when viewing them on GTS. Using this information from the crawler data, we want to add a Toast notification to GTS which is displayed when the user is viewing a non-finalized schedule. 

This component should be re-usable. Feel free to take inspiration from [react-toastify](https://www.npmjs.com/package/react-toastify) for ideas on how to make it re-usable. We would prefer not to use react-toastify.

#### Mockups 

If relevant, include mockups that the designers have made as screenshots.
```
  import React from 'react';

  import { ToastContainer, toast } from 'react-toastify';
  import 'react-toastify/dist/ReactToastify.css';
  
  function App(){
    const notify = () => toast("Wow so easy!");

    return (
      <div>
        <button onClick={notify}>Notify!</button>
        <ToastContainer />
      </div>
    );
  }
```
The component should be abstracted like the above code snippet. 

### Your Goal
- Add a new component folder inside `src/components`
- Create similar abstractions as the code snippet above

joannacheng21:
### Summary

Resolves #266 

Create reusable toast component for undecided schedules, let color, icon, message, self disappearing or not be passed in

### Checklist

- [x] Add a new component folder inside src/components
- [x] Create similar abstractions as the code snippet above

### How to Test
```
import Toast, { notifyToast } from '../Toast';

<div>
  <button type="button" onClick={(): void => notifyToast('atoast')}>
    Notify!
  </button>
  <Toast
    className="atoast"
    color="orange"
    message="Note: The final exam matrix for Fall 2023 may not be fully finalized."
    selfDisappearing={false}
  />
</div>
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ The app is a React single-page application (SPA) (built using [`create-react-app
 To implement its goal of facilitating schedule creation and class exploration, GT Scheduler stores all data **locally in cookies**. Then, it sources any relevant data at runtime from a variety of sources, such as:
 
 - The list of terms (strings like `202008`, which corresponds to Fall 2020) that have been scraped by the [Crawler application](https://api.github.com/repos/gt-scheduler/crawler/): https://gt-scheduler.github.io/crawler/index.json
-- The data for a single term, which is the full output of the Cralwer application in a single JSON file: https://gt-scheduler.github.io/crawler/202008.json
+- The data for a single term, which is the full output of the Crawler application in a single JSON file: https://gt-scheduler.github.io/crawler/202008.json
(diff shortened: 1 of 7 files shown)


yatharth-b:
It seems like the toast dissapears after 5s even when the selfDissapearing prop is set to false

yatharth-b (approved):
Looks great!

PR #292, opened 2024-02-27
Fixed Toast Notification Trigger

samarth52:
### Summary

A Toast notification for un-finalized semesters would be triggered on every re-render of the HeaderDisplay component due to the lack of memoization of props to this component. This PR fixes the problem and eliminates the usage of the deprecated `useLocalStorageNoSync` hook.


yatharth-b (approved):
LGTM! Thanks for your contribution :D

PR #297, opened 2024-03-01
Updated shared schedule deletion API calls

### LINKED ISSUE #282: [BoG S24] Delete Functions for Multiple Versions
### Problem Description

We have updated the functionality of the share schedule feature to now support sharing multiple schedules at the same time. This means that our cloud functions which are used for deleting invites from the sender's side and the receiver's side don't work anymore. 

Both the backend functions for deleting invites are in the https://github.com/gt-scheduler/firebase-conf repository.

Check out new types here - 
https://github.com/gt-scheduler/firebase-conf/blob/main/functions/utils/types.ts

### Your Goal
- In accordance with the new data type of invites, change both the delete functions to look for every single invite which has the friend, sender, and version we want to delete. Modify those invites to remove the version we are trying to delete - if this results in an empty versions array, remove the invite record entirely. We need to do this because now we are storing multiple invites from the sender to the receiver when the invites all are trying to share different sets of versions.

aeluro1:
### Updated shared schedule deletion API calls

Resolves #282


yatharth-b (approved):
LGTM!

PR #298, opened 2024-03-02
Preserve State for Compare Schedule

### LINKED ISSUE #276: [BoG S24] Preserve State for Compare Schedule
### Problem Description

Currently, the state of the compare schedule component is reset on refreshing the page. To preserve this state, we will need to make updates to [this ](https://github.com/gt-scheduler/website/blob/main/src/data/hooks/useUIStateFromStorage.ts) hook and add a new structure for keeping track of what the state of the compare schedule panel was. 

### Your Goal
- Investigate how the hook currently works.
- Add a new data type for storing the state of the compare panel, similar to how there is a data type for storing version state. 
- Create a setter for the same and update the default value of the UI state. State for how the Compare Panel works can be found here - https://github.com/gt-scheduler/website/blob/8f3616aafa8c37d5353c19941ad338d5a0fdbf69/src/components/Scheduler/index.tsx#L33 These are the only three states we need to preserve in local storage (double check this after implementing)
- Use this setter to update the local storage whenever state of the compare panel is changed. Look at how the other setters are used. For example - https://github.com/gt-scheduler/website/blob/8f3616aafa8c37d5353c19941ad338d5a0fdbf69/src/components/HeaderDisplay/index.tsx#L108
- Initalize the state of the compare panel with the data from the UI state hook

Note : it might not be necessary to have those three states for the compare panel inside `Scheduler`. These states can be moved up to the UI State hook and `ComparisonPanel` can take in value from the extracted data from localStorage.

Nathan-Papa:
Resolves #276 

Created a new hook to preserve the state of the compare schedule panel in local storage.

### How to Test

Toggle features of the compare schedule panel and refresh the page and notice they are unchanged.

### DIFF
--- a/src/components/Scheduler/index.tsx
+++ b/src/components/Scheduler/index.tsx
@@ -10,6 +10,7 @@ import {
 } from '..';
 import { OverlayCrnsContext, OverlayCrnsContextValue } from '../../contexts';
 import { DESKTOP_BREAKPOINT } from '../../constants';
+import useLocalStorageState from 'use-local-storage-state';
 import useScreenWidth from '../../hooks/useScreenWidth';
 
 /**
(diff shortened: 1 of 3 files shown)


src/data/hooks/useCompareStateFromStorage.ts
```diff
@@ -0,0 +1,75 @@
+import { useCallback } from 'react';
+import useLocalStorageState from 'use-local-storage-state';
+
+export const UI_STATE_LOCAL_STORAGE_KEY = 'ui-state';
```
yatharth-b, line 4: remove this line

src/components/Scheduler/index.tsx
```diff
@@ -30,28 +31,16 @@ export default function Scheduler(): React.ReactElement {
     [overlayCrns, setOverlayCrns]
   );
 
-  const [compare, setCompare] = useState(false);
-  const [pinnedSchedules, setPinnedSchedules] = useState<string[]>([]);
-  const [pinSelf, setPinSelf] = useState(true);
+  // const [{ currentTerm, versionStates }, setUIState] = useLocalStorageState(
```
yatharth-b, line 34: remove this comment

PR #308, opened 2024-03-10
Updated new features modal

Nathan-Papa:
### Summary

Updated the information modal to discuss the new compare schedules feature instead of recurring events.

### How to Test

Launch the page in a new window where you aren't logged in to see the new modal popup on startup.


yatharth-b (approved):
LGTM! Just changed the local storage key which tracks if the user has seen the modal

PR #317, opened 2024-03-28
Use variables for colors in invitation modal and comparison panel

joannacheng21:
### Summary

Use variables for colors in invitation modal and comparison panel

### How to Test
Check that it looks good on light and dark theme


yatharth-b (approved):
LGTM!

PR #320, opened 2024-03-31
Multiple schedule selection

### LINKED ISSUE #306: [BoG S24] Display every schedule user owns on right panel
### Problem Description

Currently only the currently selected version shows up on the right panel. We would like to show the user all versions on the right panel as well as the ability to select them for comparison.

aeluro1:
### Allow multiple user schedules to be selected during comparison

Resolves #306


yatharth-b:
Added code to make sure current version is selected by default in compare panel if it's the first time is using the panel

src/components/Calendar/index.tsx
```diff
@@ -216,6 +219,62 @@ export default function Calendar({
           meetings.push(...friendMeetings);
         })
     );
+    Object.entries(versions)
```
yatharth-b, line 222: try to combine this loop with the one, since they're the exact same. The objects being iterated over can be combined into a single array maybe.

yatharth-b (dismissed):

yatharth-b (approved):
LGTM!

Issue #265, opened 2024-01-06
.ics exported timestamps on classes in local time, not absolute time.


SpiritSeal:
**Describe the bug**
Attempted to upload class .ics file while Google Calendar and PC timezone set to MST. Uploaded class events were shifted 2 hours (i.e. class that starts at 12:30 EST is listed as starting at 12:30 MST). 

**To Reproduce**
Steps to reproduce the behavior. Example:
1. Set machine and Google Calendar timezone to MST
2. Add classes/ other events to gt-scheduler
3. Export events as .ics file
4. Import events into Google Calendar

**Expected behavior**
Events timestamps are added to Google Calendar in EST time rather than local time.

**Screenshots**
![schedule (1)](https://github.com/gt-scheduler/website/assets/45369057/90126cd2-32d2-4ed0-9bf9-5d30eae1a454)
![image](https://github.com/gt-scheduler/website/assets/45369057/bc4b7d75-9807-4c63-a155-76a5dff405cb)

hckkk:
I am also experiencing something similar, however my machine and google calendar are both set to EST but my classes are shifted by 1 hour.

SpiritSeal:
I have noticed this as well. The ical file has its timestamps early by 1 hr. I could see this being the result of some algorithm not taking into account daylight savings.

Broyojo:
ran into this too. for now, you can manually edit the .ics file

Broyojo:
see here: https://github.com/gt-scheduler/website/blob/main/src/utils/misc.tsx#L308. maybe some issue with considering EST and EDT

cgensheimer:
> see here: https://github.com/gt-scheduler/website/blob/main/src/utils/misc.tsx#L308. maybe some issue with considering EST and EDT

yep

PR #336, opened 2025-01-10
Resolve GT-Scheduler .ics export puts all classes 1 hour early

### LINKED ISSUE #334: GT-Scheduler .ics export puts all classes 1 hour early
**Describe the bug**
When my friend and I exported our respective schedules using the export "ICS Calendar file" button in the top right, we both uploaded our respective schedules to Google Calendar and the resulting schedule on Google Calendar was with every period an hour early. I used Chrome to both export and upload the file, they used Firefox for both operations. So if a class is at 12:30pm, it would be uploaded as 11:30am instead.

**To Reproduce**
Steps to reproduce the behavior. Example:
1. Go to "Export"
2. Click on "ICS (Calendar) File"
3. Go to Google Calendar
4. Click on "Settings"
5. Click on "Import & Export"
6. Upload the ICS file and click import

**Expected behavior**
The correct date and times for each class are uploaded as events to your Google Calendar

**Screenshots**
N/A

**Additional context**
I thought originally it could be a problem with timezones but at least on my Google Calendar's end my timezone and each event's timezone is right.

cgensheimer:
…e0161cbf1cd

### Summary

Resolves #334 and mentioned in #265

- Removed the static EST offset (bug exposed by Daylight Savings Time)
- Date objects should already be local timezone aware for calendar exports
- The vendor file for ICS does have a TODO for more explicit timezone handling.

### How to Test
Manual tests can be done by adjusting your machine local time and observing ICS file contents for DTSTART and DTEND.

> Expected behavior
> The correct date and times for each class are uploaded as events to your Google Calendar

### DIFF
--- a/src/utils/misc.tsx
+++ b/src/utils/misc.tsx
@@ -279,9 +279,6 @@ const getDateRange = (term: string): DateRange => {
   return { from, to };
 };
 
-// Difference between UTC and EST timezones in minutes
-export const EST_TIMEZONE_OFFSET = 240;
-
 /**
  * Exports the current schedule to a `.ics` file,
  * which allows for importing into a third-party calendar application.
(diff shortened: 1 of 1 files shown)


afazio1:
#351

PR #339, opened 2025-08-11
Attribution nits: spacing + dynamic year

afazio1:
### Spacing
Replace JSX spaces with character entities. Use the JS Date API to get the current year. 
Before:
Screenshot 2025-08-10 at 9 47 37 PM

After:
Screenshot 2025-08-10 at 9 48 23 PM

Mobile:
Screenshot 2025-08-10 at 9 47 24 PM

### How to Test
```sh
yarn start
```
See images above.


src/components/Attribution/index.tsx
```diff
@@ -11,6 +11,7 @@ import './stylesheet.scss';
 
 export default function Attribution(): React.ReactElement {
   const mobile = !useScreenWidth(DESKTOP_BREAKPOINT);
+  const year = new Date().getFullYear();
```
uma-anand, line 14: Make this a util

uma-anand: You can put it in `website/blob/main/src/utils/misc.tsx` for now but I will probably separate out utils better later

uma-anand (changes requested):
Do the util thing + can you run the install and lint commands that failed on your local and repush?

src/components/Attribution/index.tsx
afazio1: womp

uma-anand (approved):
LGTM

PR #358, opened 2025-09-21
Add support for local crawler

uma-anand:
This in combination with a PR in the crawler repo will allow someone to run a local crawler instance with the local website instance to easily test changes. Please run the website on incognito because otherwise most times it will use the cached data :(.


src/constants.ts
```diff
@@ -78,6 +78,12 @@ const BACKEND_BASE_URL = 'https://gt-scheduler.azurewebsites.net';
 const FIREBASE_PROJECT_ID = firebaseConfig.projectId || `gt-scheduler-web-dev`;
 const CLOUD_FUNCTION_BASE_URL = `https://us-east1-${FIREBASE_PROJECT_ID}.cloudfunctions.net`;
 
+// Replace URL with served url if you want to run with a local crawler
+// eg. http://localhost:8080
+const CUSTOM_CRAWLER_URL = undefined;
```
afazio1, line 83: make env var

PR #359, opened 2025-09-25
Daniel/superior seat display mechanism

### LINKED ISSUE #353: [BoG F25] Display seat information better
### Motivation
Currently the app displays seat information on hover of a course section, like this:

Image

This  display requires users to hover over a chair icon to view the number of people on the waitlist and available seats. This interaction is not intuitive and creates friction, especially on mobile devices where hover actions are inconvenient or unavailable.

### Goal
Implement this new design that is more accessible for the users.

Image

Image

We also want to prevent over-requesting the server so we want to cache the section data as much as possible.

### Current State
Right now, we fetch the seating only on a hover in the . There is also no cache or anything.

### Implementation Steps
1. **Remove existing tooltip logic:** Delete the ReactTooltip component and associated hover handlers (handleHover, hovering flag, afterShow, afterHide). Remove the sectionTooltipId usage in the seat action.

2. **Create SeatInfo component:** Create a new SeatInfo component which displays the seat information given the SeatInfoProps defined something like this:
```
type SeatInfoProps {
	isLoading: bool,
	seatInfo {
		waitlist: OccupiedInfo | None,
		inClass: OccupiedInfo | None,
	}
}

type OccupiedInfo = {
	occupied: number
	total: number
}
```
You will use the None if it is unknown.

3. Add the component to the `Section` component reflective of the design.

4. **Migrating to SWR**: Previously we were fetching the seat information only on hover. Now we would want to do a useEffect or something of that sort on render. However, it is likely that each time a call is made the section remains the same. So let's use the `useSWR` hook so we can cache the section data. This will feed the handle hover data to the SeatInfo component.

### Some Guidelines
* Make sure you have loading and unknown states for the seat information
* Test out the info on all the different colors and make sure it is visible. If you find anything inaccessible, contact Cynthia and send her a screenshot of the issue.

### Useful links
* [Section component](https://github.com/gt-scheduler/website/blob/2206fa2e58242153e0bfb773f1b905f5032d991f/src/components/Section/index.tsx#L39)
* [Figma link](https://www.figma.com/design/Qnrz8cQY7KNKqtqp89mVzm/GT-Scheduler---Fall25?node-id=2016-6214&p=f&t=3FbWDfdPWWiulGfQ-0)- you may need to ask for permissions
* [SWR](https://swr.vercel.app/)

DanielChee:
### Summary

Resolves #353 

caches seating info and removes tooltip from existing design. conditionally renders seat and waitlist info based on color of background.

### How to Test

flame my code
but my functionality is flawless
bog

### DIFF
--- a/src/components/SeatInfo/index.tsx
+++ b/src/components/SeatInfo/index.tsx
@@ -1,5 +1,8 @@
-import React, { useEffect, useState } from 'react';
+/* eslint-disable import/no-extraneous-dependencies */
+import React from 'react';
+import useSWR from 'swr';
 
+import { getContentClassName } from '../../utils/misc';
 import { Section as SectionBean } from '../../data/beans';
 
 import './stylesheet.scss';
@@ -12,6 +15,7 @@ export type OccupiedInfo = {
 export type SeatInfoProps = {
   section: SectionBean;
   term: string;
+  color: string | undefined;
 };
 
 type SeatData = {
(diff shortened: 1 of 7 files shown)


src/components/SeatInfo/index.tsx
```diff
@@ -0,0 +1,97 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
uma-anand: I'm pretty sure you can delete this line. Locally it seems to lint fine for me without the disable.

src/components/SeatInfo/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+};
+
+const fetchSeating = async (
+  section: SectionBean,
+  term: string
+): Promise<SeatData> => {
+  try {
+    const raw = await section.fetchSeating(term);
+
+    if (!raw[0] || raw[0].length < 4) {
```
uma-anand, line 36: Add a comment explaining this is the loading case/whatever other case.

src/components/SeatInfo/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+  }
+};
+
+export default function SeatInfo({
+  section,
+  term,
+  color,
+}: SeatInfoProps): React.ReactElement {
+  const { data: seatData, isLoading } = useSWR<SeatData>(
+    ['seating', section.id, term],
```
uma-anand: Instead of using the section ID as a key, use the CRN instead. Multiple classes could have the same section IDs for their sections (eg. A01). Using CRNs prevents that.

src/components/SeatInfo/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+  return (
+    <div className="SeatInfo">
+      <div className="seating-container">
+        <div className="seating">
+          <span className={`seats ${labelClass}`}>
+            Seats Filled{' '}
+            <span className={`status ${textClass}`}>
+              {isLoading
+                ? 'Loading...'
+                : seatData?.inClass
```
uma-anand: Make this `seatData?.inClass?.total` for this and the waitlist. If the total amount of seats on a waitlist is 0, we might as well just display N/A.

src/components/SeatInfo/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+    if (!raw[0] || raw[0].length < 4) {
+      return { inClass: null, waitlist: null };
+    }
+
+    const [inClassTotal, inClassOccupied, waitlistTotal, waitlistOccupied] =
+      raw[0];
+
+    return {
+      inClass: {
+        occupied: Number(inClassOccupied ?? 0),
```
uma-anand: We are repeating the same operation four times. Make this a lambda given an `OccupiedInfo` object, and then call it once for `inClass` and then the `waitlist`.

src/components/SeatInfo/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  const textClass = getContentClassName(color);
+  const labelClass =
+    textClass === 'light-content' ? 'light-label' : 'dark-label';
+
+  return (
+    <div className="SeatInfo">
+      <div className="seating-container">
+        <div className="seating">
+          <span className={`seats ${labelClass}`}>
```
uma-anand: The code for seats and waitlist is almost identical. Make this a sub component to do the display for a given OccupiedInfo | null. And call that twice, once for the seats and once for the waitlist.

src/components/SeatInfo/stylesheet.scss
```diff
@@ -0,0 +1,40 @@
...
+    display: flex;
+    flex-direction: row;
+
+    .seating {
+      display: flex;
+      flex-direction: column;
+      align-items: flex-end;
+      
+
+      .seats,
```
uma-anand: You can reduce duplication between `.seats` and `.waitlist` since they share the same rules.

src/components/SeatInfo/stylesheet.scss
```diff
@@ -0,0 +1,40 @@
...
+
+        &.light-label {
+          color: $color-neutral; // color of title on light bg
+        }
+
+        &.dark-label {
+          color: $modal-foreground-color-dark; // color of title on dark bg
+        }
+
+        .status.light-content {
```
uma-anand: The `status` classname is a little uninituitive. Try renaming some of the class names to be more clear. Eg. `seating-label`, `seating-statistics`

src/components/Section/index.tsx
```diff
@@ -99,7 +72,6 @@ export default function Section({
         },
         {
```
uma-anand: Have asked Cynthia and Spencer about this, but later, might want to delete the whole chair icon. Until we hear back, let's keep it here for now.

uma-anand (changes requested):
Great work! Left a couple comments- mostly nits. Truly a superior seating functionality.

src/components/SeatInfo/index.tsx
DanielChee: another eslint L

uma-anand (approved):

uma-anand:
I realized I forgot to mention this in my initial review, so I went ahead and made the change myself. I added functionality to extract the light label class name in the same way we do for `getContentClassName`, so it’s no longer tied to that output.

PR #360, opened 2025-09-25
Update survey banner link and token for new course details survey

uma-anand:
Update the existing survey banner to have it link to the new survey versus the old one.


afazio1 (approved):
tres chopitas
giphy

PR #363, opened 2025-09-28
Implemented location picker component

### LINKED ISSUE #354: [BoG F25] Add locations to recurring events
### Motivation
Currently we have a "Recurring Event" feature that lets users add customizable events to their schedules. This currently only allows users to add a name and date/time to the custom event.

Image

A feature request we have gotten from several survey responses is to add the ability to have locations for recurring events, and also display them in the map view. This would help people visualize their schedules better.

Current state of the map
Image

### Your Goal
1. Add the ability to add location (optional) to recurring events through a location picker.
Image

Image

2. If there is a location for the recurring event, display it in the class and the hover similar to how it is for classes right now. (There are no designs for this because it should be on the same lines as the classes but let me know if you need one.)
Current state of classes in the scheduler view:
Image

Current state of recurring events in the scheduler view:
Image

3. Add the recurring event to the maps both in the side view on the left and in the map view.

Image

Some edge cases to note:
* If the recurring event location does not fit in the current view, we do not want to change the current view. Instead the user should have to zoom out to see it. We still want to focus on the classes.
* If the recurring event location is not valid in the sense that we do not have any actual locations to match it, just mention it as not pictured. There should be a design for this, contact Cynthia if there is not.

4. **Stretch Goals**

* Keep the [course locations](https://github.com/gt-scheduler/crawler-v2/blob/088795e3830f63c9ea2f41c08c041b48cd0624e4/src/steps/parse.ts#L21) on hand and make those the first results on typing in the location picker. If it does not match one of those, then use the MapBox API.
* If someone clicks on a class/recurring event on the left sideview of the Map tab, it focuses on that location on the map if it is valid.

### Implementation Steps
1. Add the fields `where` and `location` to the [Event](https://github.com/gt-scheduler/website/blob/594219a8b0985d9d79071558ab737e0161cbf1cd/src/types.ts#L80) interface, similar to how it is for the [Meeting](https://github.com/gt-scheduler/website/blob/594219a8b0985d9d79071558ab737e0161cbf1cd/src/types.ts#L69) interface right now. I don't like that style too much, but let's try to stay consistent with the existing codebase.
Image
Image

2. Create a generic `Dropdown` component that will take in `DropdownProps` which will essentially be a searchable dropdown. Do not implement the search component here, this is supposed to be as generic as possible. The list of values it will take in for the dropdown options will be React fragments.

3. Create a `LocationPicker` component taking in `LocationPickerProps`. This is the component that given a search term will actually populate the list of locations. This should just be a wrapper on the dropdown, the sole purpose should be to make the API call to get the locations and have the CSS for each location option.

Currently we use the MapBox API for our map. We will use the same API to make the location call. Check out [this MapBox React reference](https://docs.mapbox.com/mapbox-search-js/api/react/). We will use this output to populate the autofill. There is a parameter to tune the results towards a particular location- tune it to the coordinates of the center of the map.

4. Add the location field to the [Event Add form](https://github.com/gt-scheduler/website/blob/669bba47c98cc2e78bdcac0ef8dab23902869a35/src/components/EventAdd/index.tsx#L25). On submit, update the schedule with the location and event information.

5. Update the [`EventBlocks`](https://github.com/gt-scheduler/website/blob/594219a8b0985d9d79071558ab737e0161cbf1cd/src/components/EventBlocks/index.tsx#L41) and popover to have the location too.

6. Add the recurring events to the [Map](https://github.com/gt-scheduler/website/blob/120bfd736fb0576bc002e9a7a8ea03df3676e55f/src/components/Map/index.tsx#L14) component so they are also displayed in addition to the events. You will need to change the input type to `DaySelection` to make it take a `courseEventDateMap` instead of a `courseDateMap` for the left side view. For the map view, add the recurring event locations. (Also handle behavior if we do not have the coords for a location).

Always feel free to reach out on Slack if you have any questions or need any help! This ticket is a lil big so let me know if you need an extension.

raptor1820:
### Summary

Resolves #354

This is a long commit. Everything described in the issue has been implemented except for the map auto focusing on click.

A few notes: 
1. The design and the frontend for locations not specified are not exactly 1-1 (had already worked on this before getting access to the design - can change though).
2. Minor differences in design for the location picker component.
3. There is a 300 line variable which has all GT locations which the location search can prioritize. I could not pull everything from the crawler because some non-class location don't exist on it (e.g. Bobby Dodd). It works really good though. (have to merge this constant with the course locations constant linked in the issue).
4. To implement the location picker, our mapbox usage will double. The first API call manages the search (/suggest) , and now the second API call will extract the lat and long of the specific location the user adds (/retrieve).

Screenshots
image
image
image
NYC trip appears only on zooming out manually.


afazio1:
First thing, can you rebase your changes onto main? Would like to see that it works with All days map view

src/components/DaySelection/index.tsx
```diff
@@ -139,6 +146,20 @@ export default function DaySelection({
           </div>
         );
       })}
+
+      {/* Unpictured Events Section */}
```
afazio1, line 173: Can you align this to the bottom of the container?
And style like this: 
Screenshot 2025-09-28 at 5 03 45 PM

src/components/Map/index.tsx
```diff
@@ -45,7 +49,64 @@ export default function Map(): React.ReactElement {
     });
   });
 
-  // Sort each list of course data by their times
+  // Add events to the course data and track unpictured events
+  const unpicturedEvents: Event[] = [];
```
afazio1: Can you also add virtual sections to this list. You can loop through pinnedCrns and find sections with `!location`

src/components/MapView/index.tsx
```diff
@@ -10,9 +10,11 @@ import 'mapbox-gl/dist/mapbox-gl.css';
 import './stylesheet.scss';
 
 export type MapLocation = {
-  section: string;
+  section?: string;
   id: string;
+  title?: string;
```
afazio1: Instead of making two optional types can you have Sections use the `title` prop

src/components/MapView/index.tsx
```diff
@@ -41,16 +45,25 @@ export default function MapView({
...
-  // Dark: https://api.mapbox.com/styles/v1/gt-scheduler/cktc4yzhm018w17ql65xa802o.html?fresh=true&title=copy&access_token=pk.eyJ1IjoiZ3Qtc2NoZWR1bGVyIiwiYSI6ImNrdGM0cXlqMDA0aXYyeHBma290Y2NyOTgifQ.S_A1gOu-FSQ8ywQFf2rr5A
-  // Light: https://api.mapbox.com/styles/v1/gt-scheduler/cktc4y61t018918qjynvngozg.html?fresh=true&title=copy&access_token=pk.eyJ1IjoiZ3Qtc2NoZWR1bGVyIiwiYSI6ImNrdGM0cXlqMDA0aXYyeHBma290Y2NyOTgifQ.S_A1gOu-FSQ8ywQFf2rr5A
+  // Use standard Mapbox styles instead of custom ones to avoid 403 errors
+  // Standard styles should work with any valid Mapbox token
   const [theme] = useContext(ThemeContext);
   const mapStyle =
     theme === 'dark'
-      ? 'mapbox://styles/gt-scheduler/cktc4yzhm018w17ql65xa802o' // gt-scheduler-dark
-      : 'mapbox://styles/gt-scheduler/cktc4y61t018918qjynvngozg'; // gt-scheduler-light
+      ? 'mapbox://styles/mapbox/dark-v10'
```
afazio1: is there a reason why you changed the theme?

afazio1 (changes requested):
this is really great 😊
I've done an initial review. @uma-anand will take a look tm

src/components/MapView/index.tsx
raptor1820: My bad I started messing around the config when the mapbox element wasn't loading (The API key refused to load into my local environment). I'll revert it

src/components/DaySelection/index.tsx
uma-anand, line 1: Note that we also want to display the location for recurring events in the day selection view. Figma screenshot: 
image
https://www.figma.com/design/Qnrz8cQY7KNKqtqp89mVzm/GT-Scheduler---Fall25?node-id=14975-1676&t=4RqhjGNUsIfTmZkU-0

src/components/EventAdd/index.tsx
uma-anand, line 1: Looks like when you add an event without a valid location (something like "my house" that doesn't have a valid Mapbox corresponding location) it doesn't save that location or display it in the hover. Even if we don't have a corresponding Mapnox location, we still want to save the location and display it in all the other views like we display valid locations. The only place we won't display it is the actual Map (we would still have it in the DaySelection).

src/components/DaySelection/index.tsx
uma-anand: There are designs in the Figma for this. Align to the Figma.

src/components/LocationPicker/stylesheet.scss
uma-anand, line 1: Add styling to conform with the Figma

src/components/DaySelection/index.tsx
```diff
@@ -28,10 +28,13 @@ export interface CourseDateItem {
   title: string;
   times: Period | undefined;
   daysOfWeek: string[];
+  section?: string;
+  isEvent?: boolean;
 }
```
uma-anand, line 40: Let's rename this `CourseDateItem` to `ScheduleBlockDateItem` because it's used to store both courses and events and add a type field which has an enum value `ScheduleBlockEventType` signifying if it's a course or a recurring event (this type may alreayd exist under a different name somewhere in which case use the existing type).

src/components/Dropdown/index.tsx
uma-anand, line 1: I'm not going to go too nitty gritty on this now because the focus on this sprint is the map feature. I've highlighted probably the main issues.

src/components/Dropdown/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+  }, [focusedIndex]);
+
+  // Reset option refs when options change
+  useEffect(() => {
+    optionRefs.current = optionRefs.current.slice(0, options.length);
+  }, [options]);
+
+  // Close dropdown when clicking outside
+  useEffect(() => {
+    const handleClickOutside = (event: MouseEvent): void => {
```
uma-anand: Avoid using document event listeners when you're coding in React, use React functions like `onBlur` instead

src/components/Dropdown/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+  const handleClear = (): void => {
+    if (onClear) {
+      onClear();
+    }
+    onSearchChange('');
+    setIsOpen(false);
+    setFocusedIndex(-1);
+  };
+
+  const displayValue = value && !isOpen ? String(value) : searchTerm;
```
uma-anand: Avoid casting to a string, instead you would want to do something like `value.content`

src/components/DaySelection/index.tsx
```diff
@@ -95,6 +99,7 @@ export default function DaySelection({
               getContentClassName(colorPalette[i]),
               'default'
             )}
+            // eslint-disable-next-line react/forbid-dom-props
```
uma-anand: Is this lint disable still necessary?

src/components/Event/index.tsx
```diff
@@ -82,6 +77,7 @@ export default function Event({
       {!formShown && (
         <div
           className={classes('Event', contentClassName, 'default', className)}
+          // eslint-disable-next-line react/forbid-dom-props
           style={{ backgroundColor: color }}
```
uma-anand, line 99: If the disable wasn't previously there, is it needed now?

src/components/EventAdd/index.tsx
```diff
@@ -115,6 +138,8 @@ export default function EventAdd({
                 end: parsedEnd,
               },
               days: selectedTags,
+              where: location?.where || '',
```
uma-anand: You're doing the where and location things twice. Just make a variable for the where and the location with `location?.where || ''` and `location?.location || null`

src/components/MapView/index.tsx
```diff
@@ -10,9 +10,11 @@ import 'mapbox-gl/dist/mapbox-gl.css';
 import './stylesheet.scss';
 
 export type MapLocation = {
-  section: string;
+  section?: string;
   id: string;
+  title?: string;
   coords: Location | null;
+  isEvent?: boolean;
```
uma-anand: Use the same class/event enum mentioned before

src/components/MapView/index.tsx
```diff
@@ -72,8 +85,10 @@ export default function MapView({
               <FontAwesomeIcon icon={faMapPin} className="pin-icon" />
               <div className="pin-text">
                 {coordLocations.map((coordLocation) => (
-                  <div key={coordLocation.id + coordLocation.section}>
-                    {coordLocation.id} {coordLocation.section}
+                  <div key={coordLocation.id + (coordLocation.section || '')}>
```
uma-anand: make a display section var with value `coordLocation.section || ''` and use it

src/components/Map/index.tsx
```diff
@@ -45,7 +49,64 @@ export default function Map(): React.ReactElement {
...
+    // Check if event has location data - be more defensive about type checking
+    const eventWithLocation = event as Event & {
+      where?: string;
+      location?: Location | null;
+    };
+    const eventWhere = eventWithLocation.where;
+    const eventLocation = eventWithLocation.location;
+
+    // Check if event has valid location data for map display
+    const hasValidLocation =
```
uma-anand: this has location/valid location etc logic has been similar and different over a lot of functions. try to combine them into one util function

src/components/Map/index.tsx
```diff
@@ -45,7 +49,64 @@ export default function Map(): React.ReactElement {
...
+        // Add event without location to sidebar only
+        const courseData: CombinedCourseData = {
+          id: event.id,
+          title: event.name,
+          times: event.period,
+          daysOfWeek: [...event.days],
+          coords: null,
+          isEvent: true,
+        };
+        courses.push(courseData);
```
uma-anand: most of the logic in the if/else is the same. do the combined course data with a conditional for the coords value and then push the course. then do an `if (!hasValidLocation)` where you track unpictured events

src/components/Map/index.tsx
```diff
@@ -45,7 +49,64 @@ export default function Map(): React.ReactElement {
...
+    // Check if event has valid location data for map display
+    const hasValidLocation =
+      typeof eventWhere === 'string' &&
+      eventLocation &&
+      typeof eventLocation.lat === 'number' &&
+      typeof eventLocation.long === 'number';
+
+    event.days.forEach((day) => {
+      if (!isDay(day)) return;
+      const courses = courseDateMap[day] ?? [];
```
uma-anand: doesn't this contain both courses and recurring events? rename it to schedule blocks etc

src/components/LocationPicker/index.tsx
```diff
@@ -0,0 +1,713 @@
...
+    async (query: string): Promise<void> => {
+      if (!query.trim()) {
+        setOptions([]);
+        return;
+      }
+
+      setIsLoading(true);
+
+      // First, check if the query matches any GT locations
+      const matchingGTLocations = FALLBACK_GT_LOCATIONS.filter((location) =>
```
uma-anand: This wasn't mentioned in the ticket, but can we try to do a fuzzy search here? Check out the npm fuzzy-search package

src/components/LocationPicker/index.tsx
```diff
@@ -0,0 +1,713 @@
...
+          value: {
+            where: `${location.name}, ${location.address}`,
+            location: location.coords,
+          },
+        })
+      );
+
+      try {
+        // Try MapBox Search Box API (correct format)
+        const url = new URL(
```
uma-anand: Use URL search params directly here and eveerywher else you use set
```
const params = new URLSearchParams({
  q: query,
  access_token: MAPBOX_ACCESS_TOKEN,
  session_token: `session_${Date.now()}`,
  proximity: `${GT_CAMPUS_CENTER.longitude},${GT_CAMPUS_CENTER.latitude}`,
  country: 'US',
  limit: '10',
  language: 'en',
  types: 'poi,address,place',
});
```

src/components/LocationPicker/index.tsx
```diff
@@ -0,0 +1,713 @@
...
+  };
+}
+
+interface MapBoxSearchBoxResponse {
+  suggestions: MapBoxSuggestion[];
+  attribution: string;
+}
+
+// Georgia Tech campus coordinates for proximity bias
+const GT_CAMPUS_CENTER = {
```
uma-anand: Move this constant to a constants file

src/components/LocationPicker/index.tsx
```diff
@@ -0,0 +1,713 @@
...
+  longitude: -84.3963,
+  latitude: 33.7756,
+};
+
+// Fallback GT locations with coordinates - SINGLE SOURCE OF TRUTH
+const FALLBACK_GT_LOCATIONS = [
```
uma-anand: Move this to a constants file

src/components/LocationPicker/index.tsx
```diff
@@ -0,0 +1,713 @@
...
+
+export default function LocationPicker({
+  className,
+  placeholder = 'Add location',
+  value,
+  onChange,
+  onClear,
+  disabled = false,
+}: LocationPickerProps): ReactElement {
+  const [searchTerm, setSearchTerm] = useState('');
```
uma-anand: Instead of juggling searchTerm, options, isLoading inside the component, encapsulate in a useLocationSearch hook

src/components/LocationPicker/index.tsx
uma-anand, line 1: This file is way too big, try to pull out as many functions as possible outside the component instead of inside

uma-anand: You may want to make a `utils/mapbox` file

uma-anand (changes requested):
Amazing job! All the functionality works great! I left a couple of comments. A lot of them are just along the lines of aligning with the Figma or stuff about code smells. I think what we should do is extend this ticket for half of this sprint considering there's also extensions we can still do and refactoring this can take time because it's a big ticket. Let's talk more after GBM on Tuesday.

src/components/MapView/index.tsx
raptor1820: resolved

src/components/DaySelection/index.tsx
raptor1820: Done!

src/components/Dropdown/index.tsx
raptor1820: resolved

src/components/EventAdd/index.tsx
raptor1820: added this

src/components/Event/index.tsx
raptor1820: removed in a previous commit

src/components/Dropdown/index.tsx
raptor1820: resolved

src/components/DaySelection/index.tsx
raptor1820: Added

src/components/DaySelection/index.tsx
raptor1820: Done

src/components/EventAdd/index.tsx
raptor1820: resolved

src/components/Map/index.tsx
raptor1820: fixed

src/components/EventAdd/index.tsx
raptor1820: Image

Follow up to this, I added an option at the top to use a custom location for mobile users (they can click to option). I can change the styling, if needed.  

Pressing enter on the keyboard will still work, so I can just remove the button if needed.

src/components/Map/index.tsx
raptor1820: fixed

src/components/Map/index.tsx
raptor1820: fixed

src/components/Map/index.tsx
raptor1820: refactored

src/components/LocationPicker/index.tsx
raptor1820: done

src/components/EventAdd/index.tsx
uma-anand: Looking good!

src/components/DaySelection/index.tsx
```diff
@@ -23,15 +23,24 @@ export function isDay(rawDay: string): rawDay is Day {
   }
 }
 
-export interface CourseDateItem {
+// eslint-disable-next-line no-shadow
```
uma-anand, line 27: Is this disable still needed?

src/components/DaySelection/index.tsx
```diff
@@ -23,15 +23,24 @@ export function isDay(rawDay: string): rawDay is Day {
...
+export enum ScheduleBlockEventType {
+  Course = 'course',
+  CustomEvent = 'event',
```
uma-anand: make this a custom event or something other than 'event'

src/components/LocationPicker/index.tsx
uma-anand: Looks great, thanks for cleaning this up.

src/utils/mapbox/misc.ts
uma-anand, line 1: Rename this misc following the codebase naming convention

uma-anand (changes requested):
Thanks for cleaning everything up. Looks great and clean! Comments are nits. Can you rebase and also make the destination bog sprint 3? 

Plus one thing I think we should do is instead of making where and the location separate, include an optional `where` in the Location type itself. But we can table this to a bug fix sprint or something.

src/utils/mapbox/api.ts
```diff
@@ -0,0 +1,71 @@
...
+  url.searchParams.set('access_token', accessToken);
+  url.searchParams.set('session_token', `session_${Date.now()}`);
+  url.searchParams.set(
+    'proximity',
+    `${GT_CAMPUS_CENTER.longitude},${GT_CAMPUS_CENTER.latitude}`
+  );
+  url.searchParams.set('country', 'US');
+  url.searchParams.set('limit', '10');
+  url.searchParams.set('language', 'en');
+  url.searchParams.set('types', 'poi,address,place');
```
uma-anand: Also forgot to mention just make this a group URL search param request instead of setting each thing individually.

src/components/DaySelection/index.tsx
raptor1820: Yeah this one is needed because there is a ESLint bug with enums.

raptor1820:
Changed base to [bog-f25-sprint3](https://github.com/gt-scheduler/website/tree/bog-f25-sprint3).

The only thing left to do from the ticket is autofocussing the map on clicking an event.

raptor1820:
Also, I would suggest moving the Mapbox theme info/or maybe the map itself for the MapView component into the utils/mapbox folder now that we have it.

(I edited the theme on accident while debugging some error in my branch, for example).

uma-anand (approved):

PR #373, opened 2025-10-18
Added discard button to recurring event drafts

### LINKED ISSUE #372: Add discard button to recurring event drafts
### Problem Description

We recently added a drag and drop for recurring events (#347). However, once the info has been autopopulated, there is no real way to discard it except to save the event and delete it. We want to add a discard button.

Image

#### Mockups

Image

https://www.figma.com/design/Qnrz8cQY7KNKqtqp89mVzm/GT-Scheduler---Fall25?node-id=15372-46043&p=f&t=zeWEvrg6dnZv8SZO-0

I believe this should be fairly simple so not going to add too much detail to this ticket. Checking out the PR for the original feature should tell you what you need to know. Always feel free to reach out if you have any questions!

raptor1820:
Implemented the discard button as per the issue.
image


src/components/EventAdd/index.tsx
```diff
@@ -56,6 +57,21 @@ export default function EventAdd({
...
+  useEffect(() => {
+    // For existing events, always enable discard (it deletes the event)
+    // For new events, only enable if any field has been filled
+    if (event) {
+      setDiscardDisabled(false);
+    } else {
+      const anyFieldFilled =
+        eventName.length > 0 ||
+        selectedTags.length > 0 ||
+        start !== '' ||
```
uma-anand: One thing I noticed is it doesn't enable discard if there are incomplete times filled. Eg. 
image

src/components/EventAdd/index.tsx
```diff
@@ -56,6 +57,21 @@ export default function EventAdd({
     }
   }, [eventName, selectedTags, start, end, error]);
 
+  useEffect(() => {
+    // For existing events, always enable discard (it deletes the event)
```
uma-anand: Did you verify this behavior with the PM? I would talk to Spencer and check on this decision. I think users could easily press discard while editing and delete it without expecting that behavior. The wording could also be changed to delete etc. Def check w PM.

src/components/EventAdd/index.tsx
```diff
@@ -65,6 +81,33 @@ export default function EventAdd({
     return -1; // invalid time string
   }, []);
 
+  const onDiscard = useCallback((): void => {
+    if (event) {
+      // If editing an existing event, delete it
```
uma-anand: There is the case where you could have multiple events open as a draft and in this code if you attempt to discard one what happens is it saves the other drafts while discarding that particular one. Our expected behavior would be just discarding that draft and preserving the others as drafts.
image

src/components/EventAdd/index.tsx
```diff
@@ -65,6 +81,33 @@ export default function EventAdd({
...
+      patchSchedule({
+        events: newEvents,
+        colorMap: newColorMap,
+      });
+
+      if (setFormShown) {
+        setFormShown(false);
+      }
+    } else {
+      // If creating a new event, just clear the form
```
uma-anand: You reset eventName, selectedTags, start, end, and error both on discard and on submit. Create a helper function that you call both times instead of rewriting the same code.

src/components/EventAdd/index.tsx
```diff
@@ -56,6 +57,21 @@ export default function EventAdd({
     }
   }, [eventName, selectedTags, start, end, error]);
 
+  useEffect(() => {
```
uma-anand: I realize you're using the same format as for the submit but I honestly think we can just make the discard/submit disabled just directly computed instead of complicating it with useEffect unless there's some unforseen consequence I'm missing

uma-anand (changes requested):
Thanks for working on this. I think there are a couple of things with the behavior that need some tweaking, should be mainly easy fixes.

src/components/EventAdd/index.tsx
raptor1820: I'll check with Spencer. I thought this was a general button for all drafts (whether already added or not).

spencer-banko: Discard should not delete the event when you are editing it - this is confusing. If you click discard when editing an already saved event, it should revert back to the original state of the recurring event before it was edited, and then return to the original screen before you clicked edit (recollapse the event)

src/components/EventAdd/index.tsx
raptor1820: I think removing it is the best idea. 

This ties into the bug where the discard button is not enabled even when times are partially filled. We don't track them until they are fully filled, and for the discard button, this would add extra complexity tracking hours and minutes separately.

src/components/EventAdd/index.tsx
raptor1820: fixed this

raptor1820:
You may notice I added bunch of checks for tab focus behavior and preventing race conditions on discard.

This is because of this very weird bug happening on Firefox. After discarding, approximately 1/5 times this outline remains on the calendar view, which led me to believe it was some messed up tab focus behavior or something which prevents the event blocks component from properly unmounting. Clicking anywhere on the calendar makes it disappear. 

image


I tested this today on Chrom(e/ium) and it seems to be fine.

uma-anand (approved):
Looks good! Thanks for fixing this.

PR #375, opened 2025-10-19
Added Min Travel Time Option

### LINKED ISSUE #364: [BoG F25] Add schedule combination option to minimize travel times
### Problem Description
Currently we have a number of schedule combinations listed here:
Image

These take the courses the users have added with unselected sections and gives options of sections to pick that are the "most compact" etc.

We want to add an option to minimize the travel time.

There are no mockups for this because the only UI change is going to be adding the "Least Travel Time" option to the dropdown.

Note that Ritwic was working on a [similar map ticket](https://github.com/gt-scheduler/website/pull/363), so there might be a couple conflicts. Try to use different files for utils/constants as much as possible, we can combine them later.

### Your Goal
- [ ] Copy over the latitudes longitudes used in Ritwic's ticket to `src/mapConstants`
- [ ] Use the [MapBox Matrix API](https://docs.mapbox.com/api/navigation/matrix/) once to get a matrix of distances between all the possible class locations, store this in the `src/mapConsts` file
- [ ] Create a new sorting option in [`src/data/beans/Oscar`](https://github.com/gt-scheduler/website/blob/594219a8b0985d9d79071558ab737e0161cbf1cd/src/data/beans/Oscar.ts#L177) to sort by the sum of time taken to go from each location to the next location on a particular day. This should automatically add it to the combination container without you having to specify it.

Existing sorting options (for reference):
Image

idke64:
### Summary

Resolves #364

Adds min travel time option to schedule combination dropdown.

- Used MapBox API to generate distance array.
- `GT_LOCATIONS` array has some duplicate/different names for locations.
- Uses Oliver's PR's code for consistency.

### Checklist

- [x] Copy over the latitudes longitudes used in Ritwic's ticket to `src/mapConstants`
- [x] Use the [MapBox Matrix API](https://docs.mapbox.com/api/navigation/matrix/) once to get a matrix of distances between all the possible class locations, store this in the `src/mapConsts` file
- [x] Create a new sorting option in [`src/data/beans/Oscar`](https://github.com/gt-scheduler/website/blob/594219a8b0985d9d79071558ab737e0161cbf1cd/src/data/beans/Oscar.ts#L177) to sort by the sum of time taken to go from each location to the next location on a particular day. This should automatically add it to the combination container without you having to specify it.

### How to Test

Tested by starting a development instance with `npm start` and using the option to create schedules. This assumes the MapBox distances are 100% accurate.


yarn.lock
uma-anand, line 1: Did you mean to include this file?

src/data/beans/Oscar.ts
```diff
@@ -223,6 +224,41 @@ export default class Oscar {
         const avg = sum / starts.length;
         return -avg;
       }),
+      new SortingOption('Least Travel Time', (combination) => {
+        const getTravel = (
```
uma-anand: can you make the getTravel a function in the mapconsts. i will later move the functions in that consts file to a utils file but just put it there for now.

uma-anand (changes requested):
Looks good! Very small nit but otherwise looks great!

uma-anand (approved):
LGTM! Great job.

PR #384, opened 2025-10-28
feat: show section-specific titles for 8803

### LINKED ISSUE #382: Add Section-Specific Titles for Special Topics Courses
## Problem
Currently, _Special Topics_ courses show incorrect titles. For example, all sections of CS 8803 show "Animal Interaction", but this is a section-specific title rather than the general course name. 

This happens because the crawler only stores one course title at the Course level, which gets applied to all sections regardless of their individual titles.

Related issue: [gt-scheduler/crawler-v2/45](https://github.com/gt-scheduler/crawler-v2/issues/45)

## Solution
Reference for details:
- [gt-scheduler/crawler-v2/18](https://github.com/gt-scheduler/crawler-v2/pull/18) (backend)
- [gt-scheduler/website/245](https://github.com/gt-scheduler/website/pull/245) (frontend)

Oliver-Zen:
### Summary

Resolves #382 

This PR makes the following 4 places show correct **section title** as **course name** for 8803 special topics courses:
1. SectionBlocks
2. CompareBlocks
3. Map list
4. Calendar export text

### Note: Merge Order
Merge **this** PR after the [corresponding backend PR](https://github.com/gt-scheduler/crawler-v2/pull/46). 
This PR includes a fallback, so merging out of order won’t break builds, but this order is preferred.

### Checklist

- [x] fix: Special Topics courses show incorrect titles

### How to Test

```zsh
yarn start
```
1. SectionBlocks
image

2. CompareBlocks
image

3. Map list
image

4. Calendar export text
image


src/components/CompareBlocks/index.tsx
```diff
@@ -104,7 +104,12 @@ export default function CompareBlocks({
               },
               {
                 name: 'Course Name',
-                content: section.course.title,
+                // after confirming that only new section data is served,
+                // we can remove this check for `section.sectionTitle`.
+                content:
+                  section.course.number === '8803' && section.sectionTitle
```
afazio1: can you encapsulate this into a function and put it in utils/misc.tsx since it is being repeated

afazio1 (changes requested):

src/components/CompareBlocks/index.tsx
uma-anand: I don't think we should be hardcoding 8803s (also 4803s have this same problem). I left a similar comment on the crawler PR. Can you take a look at this PR https://github.com/gt-scheduler/website/pull/245 and see if it works? Same comment for all the hardcoding.

uma-anand (changes requested):

Oliver-Zen:
### 3 Major Changes

1. No more hard coding 8803 checking: `isMultipleTopics` is computed once when the course bean is built, so every component (SectionBlocks, CompareBlocks, Map, Calendar export) can reuse the same answer without re-scanning the sections list.
2. Course / Section title fetching logic is encapsulated and put into `utils/misc.tsx`
3. Fixed map ALL day ActionRow's issue, now it shows all unique sections' titles if user registers more than one multi-topic section, instead of the first section's title

### Tests
image

- "Multiple Topics" shown as course title
- Each section block has its own course name
- Same for CompareBlocks
- Calendar export text shows section specific names

image

- Each special topic section has its own entry in ALL day ActionRow

Oliver-Zen:
Updated frontend code based on the [code changes of crawler](https://github.com/gt-scheduler/crawler-v2/pull/46).

PR #385, opened 2025-10-29
Add color palettes

### LINKED ISSUE #377: [BoG F25] Refactor Color Palettes
The new design for color palettes is [here](https://www.figma.com/design/Qnrz8cQY7KNKqtqp89mVzm/GT-Scheduler---Fall25?node-id=15240-4012&m=dev). 

Work off #371.

This design adds a dropdown in the `HeaderDisplay` component to select a color palette. 
- When a new palette is chosen, all courses in the Course Container should update their colors to use one from the palette.
	- you can do this buy updating `colorMap` in the `ScheduleData` context

We will need to now store which palette each schedule has. Update the Version3Schedule type to contain a `palette` attribute distinguishing between the different palettes available:
```ts
export interface Version3Schedule {
  desiredCourses: string[];
  pinnedCrns: string[];
  excludedCrns: string[];
  events: Event[];
  colorMap: Record<string, string>;
  sortingOptionIndex: number;
}
```

Course edit color button
- Populates with the palette selected in the Header dropdown. ie. only allow a user to choose a color within the same palette

Compare schedules color picker
- only allow a user to choose a color within the same palette

Adding new courses / recurring events:
- currently, when a new course or event is added, a random color is chosen. Refactor `getRandomColor` to take in the current `palette` for this schedule when selecting a new color.

emilyliublair:
### Summary

Resolves #377 

### How to Test
1. Courses already added change color to selected color palette
2. Default palette displayed for new schedule versions
3. getRandomColor works
4. compare schedule uses user's currently selected palette


afazio1:
One product thing:
If you have a deep palette selected already, then you change one of the colors on a class and select deep palette again it resets the colors to be incremental. Instead, if u select the palette that you already have just keep the state of the current colorMap

src/components/Course/index.tsx
```diff
@@ -35,10 +41,29 @@ export default function Course({
...
   const isSearching = Boolean(onAddCourse);
   const [
-    { oscar, desiredCourses, pinnedCrns, excludedCrns, colorMap },
+    { oscar, desiredCourses, pinnedCrns, excludedCrns, colorMap, palette },
     { patchSchedule },
   ] = useContext(ScheduleContext);
 
+  const courseRef = useRef<HTMLDivElement>(null);
+
+  useEffect((): (() => void) => {
```
afazio1: isn't this already handled by the onmouse leave event? When the mouse leaves the palette on a course, it closes the palette

src/components/Event/index.tsx
```diff
@@ -30,10 +36,26 @@ export default function Event({
   event,
 }: EventProps): React.ReactElement | null {
   const [paletteShown, setPaletteShown] = useState<boolean>(false);
-  const [{ events, colorMap }, { patchSchedule }] = useContext(ScheduleContext);
+  const [{ events, colorMap, palette }, { patchSchedule }] =
+    useContext(ScheduleContext);
   const [formShown, setFormShown] = useState<boolean>(
     Boolean(event.showEditForm)
   );
+  const eventRef = useRef<HTMLDivElement>(null);
```
afazio1: same as last comment. this should be already be handled by the on mouse leave event and can be removed

src/components/Header/index.tsx
```diff
@@ -66,6 +82,52 @@ export default function Header({
...
+        default:
+          paletteArray = DEFAULT_PALETTE.flat();
+          break;
+      }
+
+      const newColorMap: Record<string, string> = {};
+      const availableColors = [...paletteArray]; // copy to avoid mutation
+
+      // Assign each course a unique color from the palette
+      for (const courseId of Object.keys(colorMap)) {
```
afazio1: just iterate through all the courses and then have a counter `i` that indexes into the palette. So `availableColors[i % paletteArray.length]`

afazio1 (changes requested):
see comments. i fear we also broke mobile

src/components/Header/index.tsx
emilyliublair: i made it random so that it doesn't always assign the first,sec,third,etc course to the same color. i can get rid of the randomness though if you feel that it is unnecessary.

afazio1 (approved):

PR #387, opened 2025-11-01
Add TabBar and CardContainer

### LINKED ISSUE #380: Course Details Page: All Eyes on Classes
### Problem Description
We are adding a new [Course Details Page](https://www.figma.com/design/Qnrz8cQY7KNKqtqp89mVzm/GT-Scheduler---Fall25?node-id=15044-41&p=f&m=dev) to the website where users can see details about different classes they are taking. We plan on finishing this feature over the next two sprints. Your task this sprint will be to add the following components: a CardContainer, TabBar, and a CourseInfo.

Note that you will be working in combination with Lauren who will be creating the MetricsCard and ProfessorInfoCard components you will need to call (#379). You should figure out how to call her components with the right props. You can see the division of the work [here](https://excalidraw.com/#room=a958733b0e3b76baa165,mnaW1_GixIEAwpIYeLz0Xg).

### Your Goal
Create the following components.

- [ ] **CardContainer**: Create a container (the colored outline and content). It will take in a color. We should be able to put things inside the CardContainer using children like
```
<CardContainer color="blue">
        <CardContents />
</CardContainer>
```

Image

- [ ] **TabBar**: Create a tab bar component that takes in enableSelect (bool), items (TabBarItem[]), selected (TabBarItem), onSelect(key). 

If enable select is false (default), the tab bar will look like this
Image

If it is true, it will look like this
Image

- [ ] **CourseInfo**: This will be the bulk of your ticket. Create a course info component that takes in a class (like LMC 3075) and returns all the course information. This will need to use UI components that both you and Lauren have created. You will not need to populate the metrics card ratings (just place dummy values), but will need to populate the rest of the data. I believe we do not currently parse corequisites , so you may need to add that to the crawler. It should follow the same format as prereqs- let me know if you have any questions.

The course info should also have a type prop which currently can be "short" or "long".

Short:
Image

Long:
Image

For testing purposes, create a new page `/sandbox/CourseInfo` where you can place your components.

DanielChee:
### Summary

Resolves #380 

Adds a TabBar, CardContainer, and sandbox accessible by /sandbox/courseinfo

### How to Test
start website and add  /sandbox/CourseInfo to the URL


src/components/CardContainer/stylesheet.scss
```diff
@@ -0,0 +1,15 @@
+@import '../../variables';
+
+.CardContainer {
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-start;
+  align-items: stretch;
+  background-color: $content-card-background;
+  border-left: 6px solid $content-card-background;
```
uma-anand: 8px in Figma

src/components/TabBar/stylesheet.scss
```diff
@@ -0,0 +1,40 @@
+@import '../../variables';
+
+.TabBar {
+  display: flex;
+  align-items: center;
+  background-color: $tab-bar-background;
+  border-radius: 8px;
```
uma-anand: 10px in Figma

src/components/TabBar/stylesheet.scss
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    &.active {
+      background-color: rgba($color-neutral, 0.3); // adjust to match your active background tone
+      color: white;
+      font-weight: 600;
+    }
+  }
+
+  // 👇 when enableSelect is false, make all tab text white
```
uma-anand: NO EMOJIS

src/components/TabBar/stylesheet.scss
```diff
@@ -0,0 +1,40 @@
+@import '../../variables';
+
+.TabBar {
+  display: flex;
+  align-items: center;
+  background-color: $tab-bar-background;
+  border-radius: 8px;
+  padding: 4px 8px;
```
uma-anand: I feel like a lot of this isn't the actual properties in the Figma, can you double check with the designs and resubmit? Thanks

uma-anand (changes requested):
Code quality is good. Just check of the amounts for some of the paddings etc.

src/components/TabBar/stylesheet.scss
DanielChee: your welcome

src/components/TabBar/stylesheet.scss
```diff
@@ -0,0 +1,40 @@
+@import '../../variables';
+
+.TabBar {
```
uma-anand, line 9: On mobile the overflow is weird. You can take a look at the action row and make sure that this is more adaptive. Re overflow, maybe you can just hide it for now and leave a comment.

src/components/TabBar/stylesheet.scss
```diff
@@ -0,0 +1,40 @@
...
+  display: flex;
+  align-items: center;
+  background-color: $tab-bar-background;
+  border-radius: 10px;
+  padding: 2px;
+
+  .tab {
+    width: 100px;
```
uma-anand: make this max width, can you check mobile and make sure it allows at least 3 tabs for mobile

src/components/TabBar/stylesheet.scss
```diff
@@ -0,0 +1,40 @@
...
+    height: 36px;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    font-family: 'Roboto', sans-serif;
+    font-size: 14px;
```
uma-anand: we don't use px for font size in this codebase, instead we use em

uma-anand (changes requested):
Thanks for making the changes! Just have a couple of nits re smaller screen sizes for the TabBar

uma-anand:
Mobile designs are available on the Figma for your perusal

src/components/TabBar/stylesheet.scss
DanielChee: why max width? figma shows fixed width of 100

uma-anand (dismissed):
Scroll is really cool! I don't want to request more changes on this so I'll take care of removing the sandbox and maybe making the fade a hook.

PR #388, opened 2025-11-02
Added courseImport hook

raptor1820:
Implemented a hook that adds a comma separated list (non case sensitive) to the desired courses tab on the side.

Testing: Created a button to run the hook with some test courses (not committed of course, so attaching screenshots).
WhatsApp Image 2025-11-02 at 01 16 23_4c259637
WhatsApp Image 2025-11-02 at 01 16 46_7de4b689
WhatsApp Image 2025-11-02 at 01 17 11_d505a76d

### DIFF
--- a/src/hooks/useImportCourses.ts
+++ b/src/hooks/useImportCourses.ts
@@ -1,0 +1,90 @@
+import { useCallback, useContext } from 'react';
+
+import { ScheduleContext } from '../contexts';
+import { ASYNC_DELIVERY_MODE } from '../constants';
+import { getRandomColor } from '../utils/misc';
+
+/**
+ * Hook to import courses from a comma-delimited string of course names.
+ * This is useful for integrating with other GT applications like GT Roadmap.
+ *
+ * @returns A function that accepts a comma-delimited string of course names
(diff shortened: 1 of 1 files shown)


src/hooks/useImportCourses.ts
```diff
@@ -0,0 +1,90 @@
...
+        .split(',')
+        .map((name) => name.trim().toUpperCase())
+        .filter((name) => name.length > 0);
+
+      const newDesiredCourses: string[] = [];
+      const newExcludedCrns: string[] = [];
+      const newColorMap: Record<string, string> = {};
+
+      // Process each course name
+      courseNames.forEach((courseName) => {
```
afazio1, line 35: There's no need for this since we already expect this format

afazio1 (changes requested):
Looks good, just remove some logic and fix the lint error :)

src/hooks/useImportCourses.ts
raptor1820: Removed the extra parsing. 

Also maybe the local linter is not up to date? Otherwise the pre-commit hook would not have let me push my changes.

afazio1 (approved):

PR #406, opened 2025-11-11
Overflowing Location Bug

### LINKED ISSUE #383: UI Overflow on locations
I see an overflow of the location here. I think this is particularly to do with the location- maybe there is a hardcoded size somewhere for overflow? Only seems to be a problem just when the location overflows and not other fields.

Image

Check it out on the BoG sprint 3 branch

DanielChee:
### Summary

Resolves #383

Fixes overflow state for recurring events that are too small. excluded location on desktop and mobile

### How to Test
Copy Uma's exact location here:
image

make sure it doesn't overflow

play around with other times in desktop and mobile to make sure it doesn't overflow.

### DIFF
--- a/src/components/TimeBlocks/index.tsx
+++ b/src/components/TimeBlocks/index.tsx
@@ -272,11 +272,19 @@ function MeetingDayBlock({
                 ))}
             </div>
 
-            {contentBody.map((content, i) => (
-              <span className={content.className} key={`content-body-${i}`}>
-                {content.content}
-              </span>
-            ))}
+            {contentBody.map((content, i) => {
+              const name = String(content.className).toLowerCase();
+              const durationMinutes = period.end - period.start;
+              const shouldHideLocation =
+                (name === 'location' || name === 'where') &&
+                durationMinutes < 49;
+              if (shouldHideLocation) return null;
+              return (
+                <span className={content.className} key={`content-body-${i}`}>
+                  {content.content}
+                </span>
+              );
+            })}
           </div>
         )}
       </BlockElement>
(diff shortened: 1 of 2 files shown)


src/components/TimeBlocks/index.tsx
```diff
@@ -272,11 +271,21 @@ function MeetingDayBlock({
...
-              <span className={content.className} key={`content-body-${i}`}>
-                {content.content}
-              </span>
-            ))}
+            {contentBody.map((content, i) => {
+              const name = String(content.className).toLowerCase();
+              const isLocation = name === 'location' || name === 'where';
+              const isMobile =
+                typeof window !== 'undefined' &&
+                window.matchMedia('(max-width: 768px)').matches;
```
uma-anand: We want to separate business logic and styling. Don't bring in any checks here especially no magic numbers. Everything should be in the CSS.

src/components/TimeBlocks/index.tsx
```diff
@@ -272,11 +271,21 @@ function MeetingDayBlock({
...
+              if (isLocation && period.end - period.start <= 30) return null;
+              if (isMobile && isLocation && period.end - period.start <= 49)
```
uma-anand: Do all this in the CSS hiding stuff only when the rendered block height is physically too small.

You could do something like this:
```
/* If block is too short, hide location entirely */
@container (max-height: 50px) {
  .meeting-wrapper .location,
  .meeting-wrapper .where {
    display: none;
  }
```

uma-anand (changes requested):

uma-anand:
I also think that this shouldn't be allowed, where the "pm" is half covered. Either overflow should be specified or the line shouldn't be there.
image

uma-anand (approved):
LGTM! Thanks for fixing this

PR #419, opened 2025-12-28
Enable Course View Mode

uma-anand:
### Summary
As part of the new Course Details Page, we want to enable a new toggle to view course details. This PR adds the ability to toggle to a new context as well as creates the components necessary for the visual aspect of this (the button).

Currently we use a placeholder for the course details page before we go ahead with the implementation.

Note that we are still waiting on light mode designs (eta next week) so this only works on dark. 

For mobile to avoid problems with indices I also changed the way we did tabs (using indices) to using values (strings so not bulky and equality holds).

### How to Test
[Deploy Preview](https://gt-scheduler.github.io/website/pr-preview/pr-419/)

`yarn start`

image
image

Mobile
image


michaelhyi (approved):
looks good

PR #421, opened 2025-12-30
Short Course Info Details

uma-anand:
### Summary

This PR populates all the information needed for the short Course Info blob except for the rating reviews. This include Co-reqs (which are gotten with [a parallel crawler PR](https://github.com/gt-scheduler/crawler-v2/pull/53)), the description, and information about previous terms offered.

This is ready for both mobile and web (dark mode) but light mode designs are not yet available.

### How to Test
[PR Preview](https://gt-scheduler.github.io/website/pr-preview/pr-421/) (note coreqs will not work till crawler change pushed)

`yarn start`

(Run with parallel crawler)

image

A more recent class (not all terms)
image

A class with coreqs
image

Mobile
image

image

### DIFF
--- a/src/components/CourseInfo/index.tsx
+++ b/src/components/CourseInfo/index.tsx
@@ -6,7 +6,7 @@ import { CrawlerPrerequisites } from '../../types';
 import MetricsCard from '../MetricsCard';
 import TabBar from '../TabBar';
 import { ErrorWithFields, softError } from '../../log';
-import { getTermFromSemesterName } from '../../utils/semesters';
+import { getSemesterName } from '../../utils/semesters';
 
 import './stylesheet.scss';
 
@@ -59,7 +59,7 @@ export default function CourseInfo({
    * Note that course critique as of Spr2025 only has data up to Spr2024
    * This can make it incomplete (ie. From Spring 2025 to Spring 2024)
    */
-  const offered_terms = useMemo(() => {
+  const offeredTerms = useMemo(() => {
     const termInfo = course?.termInfo;
     const currentTerm = course?.term;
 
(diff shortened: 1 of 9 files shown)


src/utils/semesters.ts
```diff
@@ -41,3 +41,40 @@ export function getSemesterName(term: string): string {
   })();
   return `${semester} ${year}`;
 }
+
+/**
+ * Converts a human-facing semester name back to a term string.
+ * Example: "Fall 2021" -> "202108"
```
xHayden, line 47: ew this is a thing?

src/data/beans/Course.ts
```diff
@@ -261,7 +278,11 @@ export default class Course {
         cache = JSON.parse(rawCache) as unknown as GpaCache;
```
xHayden, line 278: is this validated at all?

src/components/CourseInfo/index.tsx
```diff
@@ -16,13 +19,65 @@ export default function CourseInfo({
...
   }, [course?.prereqs]);
   const credits = useMemo(() => course?.sections?.[0]?.credits, [course]);
 
+  /**
+   * Get the currterm + 5 terms prior to current term in course crit response
+   *
+   * Note that course critique as of Spr2025 only has data up to Spr2024
+   * This can make it incomplete (ie. From Spring 2025 to Spring 2024)
+   */
+  const offered_terms = useMemo(() => {
```
michaelhyi:
```suggestion
  const offeredTerms = useMemo(() => {
```

Slight nitpick just to keep code style consistent. not that deep if u dont change it lol

if u do be sure to change it everywhere its used

michaelhyi (approved):
tiny nitpick comment, looks great otherwise

src/utils/semesters.ts
uma-anand: Yeah becs this get is from course critique not our crawler API. We could also do it when we get data from the course critique API too which might make more sense

src/data/beans/Course.ts
uma-anand: Idts but it throws a captured error if it's wrong. Typically though it should be fine because this is our application's cache and unless someone manually edits the contents of the cache on their side it's good.

src/components/CourseInfo/index.tsx
uma-anand: Good catch I will change 😃

PR #429, opened 2026-01-29
Add light mode designs

### LINKED ISSUE #426: [BoG S26] Light Mode for Course Details
## Problem Description

Currently we have the Course Details and Section Details view all up and running for Dark Mode. Before we deploy these features we need to also get light mode working.

Light mode designs are [here](https://www.figma.com/design/Qnrz8cQY7KNKqtqp89mVzm/GT-Scheduler---Fall25?node-id=16301-4409&t=kHWcbgU05IETAO3f-0).

Don't worry about ratings but you will have to work on the View Mode tab, and the full Course Details + Section Details View.

### Development
* Work off of the `course-page-do-over` branch [here](https://github.com/gt-scheduler/website/tree/course-page-do-over). Create a new branch checking out from that one.
* In your PR, please create a detailed test plan addressing all possible scenarios like the ones in these PRs: #425 , #421 

### Implementation
* You will need to create a lot of colors probably. Note that most of these are actually just mixes of `color-neutral` in the variables.scss file. Try to reuse `color-neutral` as rgba as much as you can.
* When naming new colors, don't make them too specific like `course-details-information-table-outline-color`. Instead keep it short and snappy like `container-dark-gray` (more about the color than the specific use). Use [this chart](https://www.figma.com/design/Qnrz8cQY7KNKqtqp89mVzm/GT-Scheduler---Fall25?node-id=14913-771&t=kHWcbgU05IETAO3f-0) to motivate your color names.
* Use mixins wherever possible to set global properties instead of doing `@include light` and `@include dark` where you can
* Note that the Figma doesn't have a good color scheme for the course info card considering the background given is white. I would say just make the whole component a little darker (with all the subcomponents). You can use your discretion as to what looks good.

As always, feel free to ping me on Slack if you have any questions!

emilyliublair:
### Summary

Resolves #426 

Adds light mode designs for View Mode tab, Course Details, Section Details

### How to Test
View Mode tab:
Screenshot 2026-01-29 at 2 52 03 PM

Course Details:
Screenshot 2026-01-29 at 2 52 21 PM

Section Details:
Screenshot 2026-01-29 at 2 52 42 PM
Screenshot 2026-01-29 at 2 53 29 PM

Gradient:
Screenshot 2026-01-29 at 2 53 01 PM
Screenshot 2026-01-29 at 2 53 12 PM

Mobile:
Screenshot 2026-01-29 at 2 53 54 PM
Screenshot 2026-01-29 at 2 54 09 PM
Screenshot 2026-01-29 at 2 54 19 PM

Previous Semesters:
Screenshot 2026-01-29 at 2 55 19 PM
Screenshot 2026-01-29 at 2 55 45 PM

Dark Mode unchanged:
Screenshot 2026-01-29 at 2 56 07 PM
Screenshot 2026-01-29 at 2 56 25 PM

also fixed small bug where semester tab bar was not scrollable

### DIFF
--- a/src/components/Card/stylesheet.scss
+++ b/src/components/Card/stylesheet.scss
@@ -3,7 +3,7 @@
 .CardContainer {
   display: flex;
   flex-direction: column;
-  @include dark(background-color, $theme-dark-background-foreground);
+  @include dark(background-color, $card-dark-background);
   @include light(background-color, $card-light-background);
   border-left: 8px solid $theme-dark-background-foreground;
   border-radius: 5px;
--- a/src/components/Card/stylesheet.scss
+++ b/src/components/Card/stylesheet.scss
@@ -3,7 +3,8 @@
 .CardContainer {
   display: flex;
   flex-direction: column;
-  background-color: $theme-dark-background-foreground;
+  @include dark(background-color, $theme-dark-background-foreground);
+  @include light(background-color, $card-light-background);
   border-left: 8px solid $theme-dark-background-foreground;
   border-radius: 5px;
   padding: 20px;
--- a/src/components/CourseInfo/stylesheet.scss
+++ b/src/components/CourseInfo/stylesheet.scss
@@ -70,7 +70,6 @@
 
       .course-terms-tab-bar {
         @include dark(background-color, $buttons-navigation-hover);
-        @include light(background-color, $card-light-background);
       }
 
       .course-terms-tab-bar.enable-select {
--- a/src/components/CourseInfo/stylesheet.scss
+++ b/src/components/CourseInfo/stylesheet.scss
@@ -69,11 +69,11 @@
       gap: 20px;
 
       .course-terms-tab-bar {
-        background-color: $buttons-navigation-hover;
+        @include dark(background-color, $buttons-navigation-hover);
+        @include light(background-color, $card-light-background);
       }
 
       .course-terms-tab-bar.enable-select {
-        background-color: $theme-dark-background-foreground;
         padding: 2px;
         .tab {
           border-radius: 8px;
--- a/src/components/ProfessorInfoCard/stylesheet.scss
+++ b/src/components/ProfessorInfoCard/stylesheet.scss
@@ -7,8 +7,8 @@
   flex-direction: column;
   gap: 20px;
 
-  @include dark(background, $theme-dark-background-foreground);
-  @include light(background, $theme-light-card-background);
+  @include dark(background-color, $card-dark-background);
+  @include light(background-color, $card-light-background);
 
   .professor-header {
     padding: 0px;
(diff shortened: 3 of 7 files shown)


src/components/ProfessorInfoCard/stylesheet.scss
uma-anand, line 1: I'm not a big fan of the look of the linear gradient on the whole table box going from light (content) to dark (edges) for light mode. View the gradient more as fading to background- ie, the text becomes harder to read in the sides (more of a text color fade than a background fade). 

If that's hard, I'd prefer the text just cutting off without a gradient instead of the current look.

src/components/TabBar/stylesheet.scss
uma-anand, line 1: Can we make the term bar background color a little darker especially on the Course Info Card? I want a little more contrast. And maybe make the text darker in proportion too. I haven't double checked but I think the Tab Bar has some overwritten CSS for just the term bar (not the view mode). So that may be the right place for that.

src/components/Card/stylesheet.scss
```diff
@@ -3,7 +3,8 @@
 .CardContainer {
   display: flex;
   flex-direction: column;
-  background-color: $theme-dark-background-foreground;
+  @include dark(background-color, $theme-dark-background-foreground);
```
uma-anand: use the card dark background variable here

src/components/CourseInfo/stylesheet.scss
```diff
@@ -69,11 +69,11 @@
       gap: 20px;
 
       .course-terms-tab-bar {
-        background-color: $buttons-navigation-hover;
+        @include dark(background-color, $buttons-navigation-hover);
```
uma-anand, line 72: card-dark-background ? or no

uma-anand (changes requested):
Thanks for the detailed test plan + quick turnaround!

I've left a couple of small comments, shouldn't take too much time.

src/components/CourseInfo/stylesheet.scss
emilyliublair: they aren't the same color but i can change to $card-dark-background if you want

src/components/ProfessorInfoCard/stylesheet.scss
emilyliublair: ill just have the text cut off

src/components/TabBar/stylesheet.scss
emilyliublair: I removed the CSS in CourseInfo's stylesheet that was overwritting the TabBar CSS for light mode so that the background color is the same as that of the View Mode bar. I kept the text the same color though since it matches with the View Mode bar text color.

src/components/CourseInfo/stylesheet.scss
uma-anand: Nah i was just checking if it could be used here but this is fine

uma-anand (approved):
Thanks for the changes! Looks good.

PR #431, opened 2026-02-01
Populate course and prof metrics

### LINKED ISSUE #430: [BoG S26] Populate professor and class metrics
## Problem Description

Currently on the course details pages we display metrics. However, currently we just have sample values and not actual values tied to the backend. This Sprint, we want to change that and legitimize the data.

### Background
1. Read [this issue](https://github.com/gt-scheduler/firebase-conf/issues/28) describing the new schemas. We essentially use the new `ratings` collection to hold the ratings for all courses/professors. The `courseStats` and `professorStats` collections are aggregations of these ratings that automatically update when ratings are submitted allowing O(1) lookup.

2. Go over the `get_rating_stats` and `submit_ratings` cloud function endpoints [here](https://github.com/gt-scheduler/firebase-conf/pull/29/changes). The get endpoint will be what you will use in the frontend to get ratings. The submit ratings endpoint will be useful for you to populate data to test your implementation. You can view sample curl commands for these endpoints [here](https://github.com/gt-scheduler/firebase-conf/pull/29).

Note that the endpoints have already been deployed to the dev environment so you should be able to use them directly from there.

### Implementation Steps
1. **Add a new field in the `Course` bean called ratings.** 
      * Use the same data types and schemas as defined in firebase-conf. You can just copy-paste them.
  
2. **Create a new  `fetchRatingsInner` function within the Course bean.** 
      * Inputs should be a list of course and a list of professors 
      * You are going to want to slugify the data before putting it in the request. Normalize the course names and professor names. Only alpha numerics, separated by hyphens.
      * The `fetchGPAInner` function will be a good resource. 
      * Make sure you are validating the data while fetching it. Use the same data types and schemas as defined in firebase-conf. You can just copy-paste them. Also verify they are between 1-5.

3. **Add a `fetchRatings` function in the Course bean.**
      *  This will get the ratings for that particular class and populate the ratings field. 
      * It will call the `fetchRatingsInner` API.

4. **Add a `fetchProfessorRatings` function in the Course/Section bean (you pick).**
      * Take in a term and get all the professor ratings for the class in the term. 
      * This will probably use both the additional term info and `fetchRatingsInfo` functions if you do it in the Course bean. 
      * You can figure out the best place to store that data- whether in the `additionalTermInfo` struct or somewhere else.

5. **Call these functions in the front-end.** 
      * When loading the Course Info Page, you will need to call the `fetchRatings` for all the courses on the page. 
      * When loading the Section Info page, you will need to call the `fetchProfessorRatings` function. 
      * For the GPAs, you can use the existing GPA caches (we already fetch class GPAs and professor GPAs). I think we also fetch historical GPAs but don't really store it anywhere.
    
6. **Add caching for the ratings.** 
      * 15 minute expiry should be good for now. 
      * Take a look at the GPA cache for inspiration. 
      * Note that while testing, you can either use new incognito windows or change the cache name if you are altering the cache structure/data within.

As always, let me know if you have any questions!

emilyliublair:
### Summary

Resolves #430 

### Checklist
- [x] Added a new ratings field in the Course bean.
- [x] Created a new fetchRatingsInner function within the Course bean.
- [x] Added fetchRatings function in the Course bean.
- [x] Added fetchProfessorRatings function in the Course bean.
- [x] Called these functions in the front-end (CourseInfo). 
- [x] Added 15 minute caching for the ratings.

### How to Test
in firebase-conf (comment out auth) & deploy dev. then run:
```
curl -X POST http://localhost:5001/gt-scheduler-web-dev/us-east1/submitRatings \
  -H "Content-Type: application/x-www-form-urlencoded" \
  -d 'data={"IDToken":"abc","ratings":[{"courseId":"CS 6210","professorId":"anand-padmanabha-iyer","term":202602,"rating":4,"difficulty":4,"workload":4}]}'
```
```
curl -X POST http://localhost:5001/gt-scheduler-web-dev/us-east1/submitRatings \
  -H "Content-Type: application/x-www-form-urlencoded" \
  -d 'data={"IDToken":"abc","ratings":[{"courseId":"CS 6210","professorId":"umakishore-ramachandran","term":202602,"rating":3,"difficulty":3,"workload":3}]}'
```
Verify ratings are displayed:
Screenshot 2026-02-01 at 3 27 41 PM
Screenshot 2026-02-01 at 3 27 52 PM
Professors with no ratings:
Screenshot 2026-02-01 at 3 28 05 PM
Courses with no ratings:
Screenshot 2026-02-01 at 3 28 40 PM

Verified that deleting and re-adding a course uses cached results with 15 minute expiration:
Screenshot 2026-02-01 at 3 31 19 PM
Screenshot 2026-02-01 at 3 31 34 PM

### DIFF
--- a/src/components/CourseInfo/index.tsx
+++ b/src/components/CourseInfo/index.tsx
@@ -2,7 +2,8 @@ import React, { useContext, useEffect, useMemo, useState } from 'react';
 
 import { ScheduleContext } from '../../contexts';
 import { serializePrereqs } from '../../utils/misc';
-import { CrawlerPrerequisites } from '../../types';
+import { CourseGpa, CrawlerPrerequisites } from '../../types';
+import { slugifyProfessor } from '../../data/beans/Course';
 import MetricsCard from '../MetricsCard';
 import TabBar from '../TabBar';
 import { ErrorWithFields, softError } from '../../log';
@@ -26,19 +27,22 @@ export default function CourseInfo({
   const course = useMemo(() => oscar.findCourse(courseId), [oscar, courseId]);
 
   const [isLoaded, setIsLoaded] = useState(false);
+  const [isProfessorRatingsLoaded, setIsProfessorRatingsLoaded] =
+    useState(false);
   const [selectedTermKey, setSelectedTermKey] = useState<string | null>(null);
+  const [gpaMap, setGpaMap] = useState<CourseGpa | null>(null);
 
   useEffect(() => {
     if (course) {
-      course
-        .fetchGpa()
-        .then(() => {
+      Promise.all([course.fetchGpa(), course.fetchRatings()])
+        .then(([gpaData]) => {
+          setGpaMap(gpaData);
           setIsLoaded(true);
         })
         .catch((err) => {
           softError(
             new ErrorWithFields({
-              message: 'error fetching course GPA',
+              message: 'error fetching course data',
               source: err,
               fields: {
                 courseId,
@@ -96,13 +100,31 @@ export default function CourseInfo({
 
   useEffect(() => {
     if (enableTermSelect) {
+      if (course) {
+        Promise.all([course.fetchProfessorRatings(course.term)])
+          .then(() => {
+            setIsProfessorRatingsLoaded(true);
+          })
+          .catch((err) => {
+            softError(
+              new ErrorWithFields({
+                message: 'error fetching professor ratings',
+                source: err,
+                fields: {
+                  courseId,
+                  term: course.term,
+                },
+              })
+            );
+          });
+      }
       if (offeredTerms.length > 0 && !selectedTermKey) {
         setSelectedTermKey(
           offeredTerms && offeredTerms[0] ? offeredTerms[0] : null
         );
       }
     }
-  }, [enableTermSelect, offeredTerms, selectedTermKey]);
+  }, [enableTermSelect, course, offeredTerms, selectedTermKey]);
 
   const instructorsForSelectedTerm = useMemo(() => {
     if (!course || !selectedTermKey) return [];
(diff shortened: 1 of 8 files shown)


src/components/CourseInfo/index.tsx
```diff
@@ -102,7 +125,71 @@ export default function CourseInfo({
...
+            })
+          );
+        });
+    }
+  }, [selectedTermKey, course, courseId]);
+
+  const formatValue = (val: number, decimals: number): string =>
+    Number.isInteger(val) ? val.toString() : val.toFixed(decimals);
+
+  const metrics = useMemo(() => {
```
uma-anand, line 149: Instead of displaying a hyphen can we display N/A and remove the unit? it might do to implement this in the metric card. ie, value ?? "N/A"

src/data/beans/Course.ts
uma-anand, line 1: Can you move the types that you copied from the firebase-conf file to `website/src/data/types`?

src/data/beans/Course.ts
```diff
@@ -43,6 +50,42 @@ interface SectionGroup {
...
+export type RatingStatsResponse = z.infer<typeof RatingStatsResponseSchema>;
+
+export const slugifyCourse = (courseId: string): string => {
+  const [subject, number] = courseId.split(' ');
+  if (!subject || !number) return courseId;
+  const cleanNumber = number.replace(/\D/g, '');
+  return `${subject} ${cleanNumber}`;
+};
+
+export const slugifyProfessor = (name: string): string => {
```
uma-anand: Name this function `slugify` and move it to `src/utils/misc.tsx`.

src/data/beans/Course.ts
```diff
@@ -43,6 +50,42 @@ interface SectionGroup {
...
+}
+
+const RatingStatsResponseSchema = z.object({
+  courses: z.record(z.string(), NormalizedStatSchema.nullable()),
+  professors: z.record(z.string(), NormalizedStatSchema.nullable()),
+});
+
+export type RatingStatsResponse = z.infer<typeof RatingStatsResponseSchema>;
+
+export const slugifyCourse = (courseId: string): string => {
```
uma-anand: Just use the `normalizeCourseName` function already present in `src/utils/misc.tsx`. Also if there is any chance a section name can slip in that function, add a flag to make sure it doesn't.

src/data/beans/Course.ts
```diff
@@ -333,6 +383,268 @@ export default class Course {
...
+
+  private async fetchRatingsInner(
+    courses: string[],
+    professors: string[]
+  ): Promise<RatingStatsResponse | null> {
+    const normalizedCourses = courses.map((c) => slugifyCourse(c));
+    const normalizedProfessors = professors.map((p) => slugifyProfessor(p));
+
+    // TODO: update URL when deployed
+    const url = `http://localhost:5001/gt-scheduler-web-dev/us-east1/getRatingStats`;
```
uma-anand: Place the URL at the top of the file as a constant and write a small comment explaining what the cache contains.

src/data/beans/Course.ts
```diff
@@ -333,6 +383,268 @@ export default class Course {
...
+              url,
+            },
+          })
+        );
+      }
+
+      return null;
+    }
+
+    // Validate the response data
```
uma-anand: Make this whole validation a reusable function in `misc` because you will probably reuse it next sprint while submitting ratings

src/data/beans/Course.ts
```diff
@@ -333,6 +383,268 @@ export default class Course {
...
+            courses: normalizedCourses,
+            professors: normalizedProfessors,
+          },
+        })
+      );
+      return null;
+    }
+  }
+
+  async fetchProfessorRatings(
```
uma-anand, line 489: Instead of this "all cached or nothing" approach, let's do selective fetching. In the current approach, if we already fetched "Anand Padmanabha Iyer" and "Umakishore Ramachandran" from CS6210, and we are trying to fetch CS3210 ("Anand Padmanabha Iyer" and "Alexey Tumanov"), we do not want to refetch "Anand Padmanabha Iyer. 

For that purpose, use the following approach:
1. For each professor:
- if cache entry exists and is valid → use it
- else → mark it as missing / expired
2. For all marked:
- fetch only those professors ratings (all together)
- merge results into cache + in-memory data
3. Return results for the term from the merged data

src/data/beans/Course.ts
uma-anand, line 1: You don't have to do this now but a future improvement would be creating a cache interface with getter or setter since we spend so many lines of code on cache operations. Since this would involve refactoring existing code too, let's make it a potential refactor for a sprint day etc

src/components/CourseInfo/index.tsx
```diff
@@ -26,19 +27,41 @@ export default function CourseInfo({
...
+              message: 'error fetching course gpa',
+              source: err,
+              fields: {
+                courseId,
+                term: course.term,
+              },
+            })
+          );
+        });
+      course
```
uma-anand: Instead of having separate errors for both (bloat, we also don't need an error that specific), we can just use one effect + error for both.

Something like this:

```useEffect(() => {
  if (!course) return;

  (async () => {
    try {
      const gpa = await course.fetchGpa();
      setGpaMap(gpa);
      await course.fetchRatings();
    } catch (err) {
      softError(
        new ErrorWithFields({
          message: 'error fetching course metrics',
          source: err,
          fields: { courseId },
        })
      );
    }
  })();
}, [course, courseId]);

File: src/components/CourseInfo/index.tsx [hunk 8]
```diff
@@ -111,30 +198,69 @@ export default function CourseInfo({
...
+
+    instructorsForSelectedTerm.forEach((professorName) => {
+      const profGpa = gpaMap?.[professorName];
+      const slugifiedName = slugifyProfessor(professorName);
+
+      if (!isProfessorRatingsLoaded || !course?.professorRatings) {
+        metricsMap[slugifiedName] = [
+          { label: 'Overall Rating', value: '—' },
+          {
+            label: 'Course GPA',
```
uma-anand: I think this should be something other than Course GPA? I haven't checked the Figma but shouldn't this be Prof GPA or smthg?

src/components/CourseInfo/index.tsx
uma-anand, line 1: The Prof and Course Metric code is a little messy (the parts where metrics are constructed with labels etc.). Can you try to clean it up in general? Also it may help make it so that the prof info card handles the labels for the prof data.

src/components/CourseInfo/index.tsx
```diff
@@ -202,7 +328,14 @@ export default function CourseInfo({
                   <ProfessorInfoCard
                     key={instructorName}
                     professorName={instructorName}
-                    professorMetrics={metrics}
+                    professorMetrics={
+                      professorMetricsMap[slugifyProfessor(instructorName)] ?? [
+                        { label: 'Overall Rating', value: '—' },
+                        { label: 'Course GPA', value: '—' },
```
uma-anand: prof gpa?

uma-anand (changes requested):
Thanks for the fast turnover! Sorry for the delayed review, I missed the email :(

I haven't tested it other than the GPA stuff. I'll wait for the next ticket when everything is deployed in dev. There are quite a few requested changes but hopefully most don't take too long.

src/components/CourseInfo/index.tsx
emilyliublair: The Figma says Course GPA, but it also has the hrs/wk labeled as Course GPA as well. I can update to Professor GPA if you want. (ref: https://www.figma.com/design/Qnrz8cQY7KNKqtqp89mVzm/GT-Scheduler---Fall25?node-id=15303-15423&t=IN8zakHsr90WDg7o-0)

src/components/CourseInfo/index.tsx
```diff
@@ -35,33 +32,16 @@ export default function CourseInfo({
 
   useEffect(() => {
     if (course) {
-      course
-        .fetchGpa()
-        .then((gpaData) => {
+      Promise.all([course.fetchGpa(), course.fetchRatings()])
```
uma-anand: Let's not do Promise.all() because we still want the GPA stuff to finish even if the ratings doesn't and vice versa.

uma-anand: Hmm I guess let's just keep it as it was before then with the two useEffects if we don't want to make it run sequentially. sorry bout that

src/components/CourseInfo/index.tsx
```diff
@@ -149,43 +129,38 @@ export default function CourseInfo({
     }
   }, [selectedTermKey, course, courseId]);
 
-  const formatValue = (val: number, decimals: number): string =>
-    Number.isInteger(val) ? val.toString() : val.toFixed(decimals);
-
   const metrics = useMemo(() => {
+    const courseRatings = course?.ratings;
+    const courseGpa = gpaMap?.averageGpa;
```
uma-anand, line 151: Add the professor ratings as a const here to keep it consistent

uma-anand (changes requested):
This is pretty much good to go, just two very small changes and we shld be good

uma-anand (approved):
Thanks for all the changes! Sorry about the confusion!

Issue #417, opened 2025-12-13
Lookup and Add Classes by CRN


Arjun-Dulam:
**The feature**
Adding on to the option of searching up classes via course prefix and number, the proposed new feature is to add an input field that will allow a user to add a course and section using its CRN alone. This would remove the hassle of first selecting a course, then a section in order to add it to the schedule.

jayvenn21:
I'm implementing this now and will open a PR shortly.

PR #434, opened 2026-02-08
feat: add ability to add classes by CRN

jayvenn21:
This PR adds support for adding a class directly by CRN.

Users can enter a CRN to resolve the corresponding section and add it to their schedule without first selecting the course and section manually.

The implementation reuses existing section data and schedule mutation logic, and includes basic validation and error handling for invalid or duplicate CRNs.

No existing behavior is changed.
#417


sweezyio:
This would appear to also close #122

PR #443, opened 2026-03-10
Add Major Restrictions

### LINKED ISSUE #415: [BoG S26] Rearrange major restrictions crawler output to use indices
## Problem Statement
The current major restrictions crawler output is too long. 

Image

We have a data structure this big for every section which is bigger than the entire rest of the section information. This makes the data a lot harder and slower to load (take a look at the difference [before ](https://gt-scheduler.github.io/crawler-v2/202302.json) and [after](https://gt-scheduler.github.io/crawler-v2/202602.json)) There is also no consistency with the rest of the crawler data formatting.

### Expected Crawler Output
There are a couple ways to restructure the output so it becomes smaller. All these ways are already employed in other parts of the application. This includes a) Structuring as tuples / tuple of tuples instead of dictionaries to avoid needing to print the keys b) While crawling, compile a list of all the possible levels/majors and index into those arrays as needed. 

You can see examples of these in the caches section of the crawler output. 

Image

A sample way of structuring the tuple could be:
```
restrictions_list = [allowed restrictions list, disallowed restrictions list]
restriction = [category-index, category-value-index]
```

So for the example, we would want something like this: 
```
..., [[1 (stands for level category of restrictions}, 0 (stands for GT Atlanta campus)], ... (disallowed restrictions)]
...
cache: {
    "restrictions": ["Campus", "Level", ...], // new
    "campus": ["Georgia Tech-Atlanta *", ...] //already existing
    ...
}
```

### Expected UI Output
Refer Figma.

emilyliublair:
### Summary

Resolves #415

### How to Test
- `yarn serve` in `emily/rearrange_crawler_output` branch of crawler-v2
- modify CRAWLER_BASE_URL to point to local crawler

Use CS3001 section A as example:
Screenshot 2026-03-10 at 7 52 51 PM

1. Opening from Course Details page
Screenshot 2026-03-10 at 7 47 40 PM
Screenshot 2026-03-10 at 7 48 03 PM

2. Opening from CourseInfoModal
Screenshot 2026-03-10 at 7 48 21 PM

3. Dark Theme 
Screenshot 2026-03-10 at 7 50 24 PM
Screenshot 2026-03-10 at 7 50 42 PM

4. No ellipses
Screenshot 2026-03-10 at 7 58 01 PM

5. Mobile View
Screenshot 2026-03-11 at 11 02 36 AM
Screenshot 2026-03-11 at 11 02 54 AM

### DIFF
--- a/src/components/CourseInfo/index.tsx
+++ b/src/components/CourseInfo/index.tsx
@@ -11,6 +11,7 @@ import { ErrorWithFields, softError } from '../../log';
 import { getSemesterName } from '../../utils/semesters';
 import ProfessorInfoCard from '../ProfessorInfoCard';
 import Button from '../Button';
+import { MajorRestrictionsView } from '../MajorRestrictionsView';
 
 import './stylesheet.scss';
 
@@ -73,7 +74,6 @@ function CourseHeader({
   );
 }
 
-// Need to create course info short and course info long
 export default function CourseInfo({
   courseId,
   enableTermSelect = false,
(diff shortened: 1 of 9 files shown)


uma-anand (approved):
looks good; i made a few changes to styling but that's pretty much it

